PR #2, opened 2013-02-06
NullPointerException in ckeditor widget after ajax update

TobiasWebelsiep:
The form content mask contains a ckeditor. The mask will be replaces by a new one containing a new instance of ckeditor (forward navigation in our application). If we navigate back and typing changes in first ckeditor instance the console shows NPE.

ckeditor_npe


tandraschko:
Thanks for the patch.

PR #3, opened 2013-02-25
Add line&column support to CodeMirror completion event

planetsizebrain:
Together with a pull request on the resources-codemirror module these changes add line&column support to the CodeMirror completion event.

I've added this because I'm working on a Java compiler based autocompletion module (that needs the cursor location) that can be used in combination with the CodeMirror module to create a sort of simple Java IDE functionality that can be integrated into JSF webapps.

### DIFF
--- a/src/main/java/org/primefaces/extensions/component/codemirror/CodeMirror.java
+++ b/src/main/java/org/primefaces/extensions/component/codemirror/CodeMirror.java
@@ -449,7 +449,7 @@ public class CodeMirror extends HtmlInputTextarea implements ClientBehaviorHolde
 
 			suggestions = (List<String>) completeMethod.invoke(
 					facesContext.getELContext(),
-					new Object[] { completeEvent.getToken(), completeEvent.getContext()});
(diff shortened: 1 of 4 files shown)


ova2:
Thomas, a task for you :-)

tandraschko:
Thanks, i will check it tomorrow.
Just one question: Wouldn't it be better the pass the complete "CompleteEvent" object to the listener method?

planetsizebrain:
1 _CompleteEvent_ object is indeed better and cleaner than 4 separate values, so I changed it and committed it.

tandraschko:
Would you please test my commited version?
I didn't like the additional cursor class and the exception handling.

planetsizebrain:
I tested your committed version and it works OK. 

I just added the cursor class and exception handling to handle cases where the wrapped CodeMirror javascript component would pass possibly unexpected non-integer values for the line & column parameters which would throw a NumberFormatException with unexpected results.

tandraschko:
IMO complete should never return non-integer values as you always are inside a row/column when a complete event will be fired

PR #4, opened 2013-06-27
Timeline events are offset (1)

brooksie:
Now respects daylight saving correctly - but the timeline ALWAYS displays using local (browser) time zone, regardless of the "timeZone" parameter in the JSF tag. It could be resolved by adjusting the UTC time, but the problem is I don't know how much to adjust the time by. I can see three possible solutions: (1) Identify what timeZone the browser is using; (2) make the timeline always display based on UTC time, rather than local time; or (3) make the javascript take account of the specified timezone when displaying (but from my limited research I don't see Javascript Dates being bale to do this). Can anyone contribute to solving this please?

### DIFF
--- a/src/main/java/org/primefaces/extensions/util/DateUtils.java
+++ b/src/main/java/org/primefaces/extensions/util/DateUtils.java
@@ -47,7 +47,7 @@ public class DateUtils {
 	public static Date toUtcDate(Calendar calendar, TimeZone localTimeZone, Date localDate) {
 		int offsetFromUTC = localTimeZone.getOffset(localDate.getTime()) * (-1);
 		calendar.setTime(localDate);
-		calendar.add(Calendar.MILLISECOND, offsetFromUTC);
+		//calendar.add(Calendar.MILLISECOND, offsetFromUTC);
 
 		return calendar.getTime();
 	}
@@ -56,7 +56,7 @@ public class DateUtils {
 	public static long toLocalDate(Calendar calendar, TimeZone localTimeZone, Date utcDate) {
 		int offsetFromUTC = localTimeZone.getOffset(utcDate.getTime());
 		calendar.setTime(utcDate);
-		calendar.add(Calendar.MILLISECOND, offsetFromUTC);
+		//calendar.add(Calendar.MILLISECOND, offsetFromUTC);
 
 		return calendar.getTimeInMillis();
 	}


ova2:
"timeZone" attribute was working in my last tests and tests of other users. I can test again. Are you sure you set the correct "timeZone"? This is either a TimeZone object or a String representing time zone Id.

brooksie:
Please close this pull request (don't merge).

brooksie:
Oleg
This is a long note, and a design discussion - is there somewhere more appropriate to post this?

I agree my last suggested code change was incorrect. After much investigation, I think there are two separate issues with the way timeline displays the events, both related to the "timeZone" parameter. The code today in PFE is OK, but doesn't solve the issues. 

The Basics: there are two significant timezones that need to be considered - the TZ that your PC / browser is using (I will use the term browserTZ) and the TZ that you want to use to display the timeline (targetTZ). Today we only know the timeZone attribute, which is the targetTZ. 

Issue 1: The time line doesn't always display correctly if targetTZ is different from browserTZ when daylight saving starts or ends.

The timeline is rendered using Javascript, and Javascript date objects only supports UTC time and local time (browserTZ). When targetTZ is different and daylight saving time starts/ends at a different time to the browserTZ, there is an error in the display of the timezone.

For example: If your browser is running in Paris time, DST started on 31-March-13 @ 01:00 UTC (2am Paris).  When the timeline displays, it correctly shows 00:00, 01:00, 03:00. This is regardless of the targetTZ. So if your targetTZ is Chicago, which is already on DST, the 2am hour is missing from the timeline display, because the rendering is based on the browserTZ.

This is not a huge issue (it is one hour twice a year), but it is a fundamental issue because there is no easy way to resolve it. Either you have to build your own replacement for the JS Date object that comprehends other timezones, or re-architect the timeline to be rendered from the server. My view is you can live with this.

Issue 2: The event times are not translated to targetTZ correctly

For example: Consider a 12 hour event starting at 23:00 UTC on 30-March-13 and ending at 11:00 UTC. Display the timeline using a targetTZ = Chicago, but a browserTZ set for Paris. The correct time conversions are:
30-March @ 23:00 UTC = 31-Mar @ 00:00 Paris (UTC+1) = 30-Mar @ 18:00 Chicago (UTC-5). 

31-March @ 11:00 UTC = 31-Mar @ 13:00 Paris (UTC+2) = 31-Mar @ 06:00 Chicago (UTC-5)
Chicago started DST earlier in March, so is UTC-5 rather than UTC-6; Paris started it at 2AM on 31-March going from UTC+1 to UTC+2.

What is displayed today by timeline is an event from 19:00 --> 08:00. What should be displayed is 18:00 --> 06:00. 

The further away your browserTZ is from UTC, the worse the errot gets. Even when browserTZ = targetTZ = Paris, the timeline displays 01:00 --> 15:00 instead of 00:00 --> 13:00.

The issue is there isn't enough information to do the correct adjustment. You need both browserTZ and targetTZ. Today you only have targetTZ. You make the best use you can of targetTZ, but it isn't enough. 

My proposal: On the assumption that the timeZone attribute means targetTZ (which defaults to the server's TZ if not specified), and add a new attribute "browserTZ", which defaults to timeZone (=targetTZ) if not specified... thus maintaining some sort of backward compatibility. Then redefine DateUtils to do the UTC adjustment as
    localDate = UTC + targetTZ offset - browserTZ offset

The
 conversion above will become 23:00 + (- 5
 hrs) - (+ 1 hr) = 17:00 UTC, which then gets rendered as 18:00 on the timeline (UTC+1)
and 11:00 UTC + (-5 hours) - (+2 hours) = 04:00UTC, which is rendered as 06:00  (UTC+2).
End result - it correctly displays as 18:00 --> 06:00.

A better solution would be to detect the browser's TZ, but from my reading (e.g. http://blog.redfin.com/devblog/2007/08/getting_the_time_zone_from_a_web_browser.html#.UgK1lm0yKpo) it's not easy to do and in any case we need it at the point of rendering the page, not when the javascript is executed. Does anyone know how to solve this?

An alternative proposal would be to get the javascript to display the timeline in UTC, rather than local time. This would solve the problem of knowing what TZ the browser was using, but the timeline wouldn't display correctly
 on DST changes (similar to issue 1). Do you / anyone know if this is an easy solution?

Thoughts? I have a mocked up version of the revised timeline with the extra attribute if you want me to send a pull request.

Regards, 

    Steve.

ova2:
Yeah, browser TZ is not supported because it requires an extra big JS library like this one https://github.com/mde/timezone-js

PR #5, opened 2013-10-06
Corrects "Timeline events are offset" issue

brooksie:
Update to timeline to correctly respect timezones and avoid offset problems. Will display events correctly positioned in alternate target timeZones. Requires a new parameter "browserTZ" to do the translation. BrowserTZ should be specified whenever Timezone is used. If not specified, defaults to same as TimeZone. Apologies but the changes were built against a 0.7.1 baseline.

### DIFF
--- a/src/main/java/org/primefaces/extensions/component/timeline/DefaultTimelineUpdater.java
+++ b/src/main/java/org/primefaces/extensions/component/timeline/DefaultTimelineUpdater.java
@@ -21,7 +21,6 @@ package org.primefaces.extensions.component.timeline;
 import java.io.IOException;
 import java.io.Serializable;
 import java.util.ArrayList;
-import java.util.Calendar;
 import java.util.List;
 import java.util.TimeZone;
 import java.util.logging.Level;
@@ -91,10 +90,12 @@ public class DefaultTimelineUpdater extends TimelineUpdater implements PhaseList
 		crudOperationDatas.add(new CrudOperationData(CrudOperation.CLEAR));
 	}
 
+        @Override
 	public PhaseId getPhaseId() {
 		return PhaseId.RENDER_RESPONSE;
 	}
 
+        @Override
 	public void beforePhase(PhaseEvent event) {
 		if (crudOperationDatas == null) {
 			return;
@@ -109,8 +110,8 @@ public class DefaultTimelineUpdater extends TimelineUpdater implements PhaseList
 		TimelineRenderer timelineRenderer =
 		    (TimelineRenderer) fc.getRenderKit().getRenderer(Timeline.COMPONENT_FAMILY, Timeline.DEFAULT_RENDERER);
 
-		TimeZone timeZone = ComponentUtils.resolveTimeZone(timeline.getTimeZone());
-		Calendar calendar = Calendar.getInstance(timeZone);
+		TimeZone targetTZ = ComponentUtils.resolveTimeZone(timeline.getTimeZone());
+		TimeZone browserTZ = ComponentUtils.resolveTimeZone(timeline.getBrowserTZ(), targetTZ);
 
 		try {
 			for (CrudOperationData crudOperationData : crudOperationDatas) {
@@ -120,7 +121,7 @@ public class DefaultTimelineUpdater extends TimelineUpdater implements PhaseList
 					sb.append(";");
 					sb.append(widgetVar);
 					sb.append(".addEvent(");
-					sb.append(timelineRenderer.encodeEvent(fc, fsw, fswHtml, timeline, calendar, timeZone,
+					sb.append(timelineRenderer.encodeEvent(fc, fsw, fswHtml, timeline, browserTZ, targetTZ,
 					                                       crudOperationData.getEvent()));
 					sb.append(")");
 					break;
@@ -132,7 +133,7 @@ public class DefaultTimelineUpdater extends TimelineUpdater implements PhaseList
 					sb.append(".changeEvent(");
 					sb.append(crudOperationData.getIndex());
 					sb.append(",");
-					sb.append(timelineRenderer.encodeEvent(fc, fsw, fswHtml, timeline, calendar, timeZone,
+					sb.append(timelineRenderer.encodeEvent(fc, fsw, fswHtml, timeline, browserTZ, targetTZ,
 					                                       crudOperationData.getEvent()));
 					sb.append(")");
 					break;
@@ -171,6 +172,7 @@ public class DefaultTimelineUpdater extends TimelineUpdater implements PhaseList
 		}
 	}
 
+        @Override
 	public void afterPhase(PhaseEvent event) {
 		// NOOP.
 	}
--- a/src/main/java/org/primefaces/extensions/component/timeline/Timeline.java
+++ b/src/main/java/org/primefaces/extensions/component/timeline/Timeline.java
@@ -19,7 +19,6 @@ package org.primefaces.extensions.component.timeline;
 
 import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.Calendar;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Date;
@@ -87,6 +86,7 @@ public class Timeline extends UIComponentBase implements Widget, ClientBehaviorH
 		var,
 		locale,
 		timeZone,
+		browserTZ,
 		style,
 		styleClass,
 		height,
@@ -181,6 +181,14 @@ public class Timeline extends UIComponentBase implements Widget, ClientBehaviorH
 
 	public void setTimeZone(Object timeZone) {
 		setAttribute(PropertyKeys.timeZone, timeZone);
+	}
+
+	public Object getBrowserTZ() {
+		return getStateHelper().eval(PropertyKeys.browserTZ, null);
+	}
+
+	public void setBrowserTZ(Object timeZone) {
+		setAttribute(PropertyKeys.browserTZ, timeZone);
 	}
 
 	public String getStyle() {
@@ -448,12 +456,12 @@ public class Timeline extends UIComponentBase implements Widget, ClientBehaviorH
 
 			if ("add".equals(eventName)) {
 				// preset start / end date and the group
-				Calendar calendar = Calendar.getInstance(TimeZone.getTimeZone("UTC"));
-				TimeZone timeZone = ComponentUtils.resolveTimeZone(getTimeZone());
+				TimeZone targetTZ = ComponentUtils.resolveTimeZone(getTimeZone());
+				TimeZone browserTZ = ComponentUtils.resolveTimeZone(getBrowserTZ(), targetTZ);
 				TimelineAddEvent te =
 				    new TimelineAddEvent(this, behaviorEvent.getBehavior(),
-				                         DateUtils.toUtcDate(calendar, timeZone, params.get(clientId + "_startDate")),
-				                         DateUtils.toUtcDate(calendar, timeZone, params.get(clientId + "_endDate")),
+				                         DateUtils.toUtcDate(browserTZ, targetTZ, params.get(clientId + "_startDate")),
+				                         DateUtils.toUtcDate(browserTZ, targetTZ, params.get(clientId + "_endDate")),
 				                         params.get(clientId + "_group"));
 				te.setPhaseId(behaviorEvent.getPhaseId());
 				super.queueEvent(te);
@@ -470,10 +478,10 @@ public class Timeline extends UIComponentBase implements Widget, ClientBehaviorH
 					clonedEvent.setStyleClass(timelineEvent.getStyleClass());
 
 					// update start / end date and the group
-					Calendar calendar = Calendar.getInstance(TimeZone.getTimeZone("UTC"));
-					TimeZone timeZone = ComponentUtils.resolveTimeZone(getTimeZone());
-					clonedEvent.setStartDate(DateUtils.toUtcDate(calendar, timeZone, params.get(clientId + "_startDate")));
-					clonedEvent.setEndDate(DateUtils.toUtcDate(calendar, timeZone, params.get(clientId + "_endDate")));
+					TimeZone targetTZ = ComponentUtils.resolveTimeZone(getTimeZone());
+					TimeZone browserTZ = ComponentUtils.resolveTimeZone(getBrowserTZ(), targetTZ);
+					clonedEvent.setStartDate(DateUtils.toUtcDate(browserTZ, targetTZ, params.get(clientId + "_startDate")));
+					clonedEvent.setEndDate(DateUtils.toUtcDate(browserTZ, targetTZ, params.get(clientId + "_endDate")));
 					clonedEvent.setGroup(params.get(clientId + "_group"));
 				}
 
@@ -510,24 +518,24 @@ public class Timeline extends UIComponentBase implements Widget, ClientBehaviorH
 				return;
 			} else if ("rangechange".equals(eventName)) {
 				// get start / end date
-				Calendar calendar = Calendar.getInstance(TimeZone.getTimeZone("UTC"));
-				TimeZone timeZone = ComponentUtils.resolveTimeZone(getTimeZone());
+				TimeZone targetTZ = ComponentUtils.resolveTimeZone(getTimeZone());
+				TimeZone browserTZ = ComponentUtils.resolveTimeZone(getBrowserTZ(), targetTZ);
 				TimelineRangeEvent te =
 				    new TimelineRangeEvent(this, behaviorEvent.getBehavior(),
-				                           DateUtils.toUtcDate(calendar, timeZone, params.get(clientId + "_startDate")),
-				                           DateUtils.toUtcDate(calendar, timeZone, params.get(clientId + "_endDate")));
+				                           DateUtils.toUtcDate(browserTZ, targetTZ, params.get(clientId + "_startDate")),
+				                           DateUtils.toUtcDate(browserTZ, targetTZ, params.get(clientId + "_endDate")));
 				te.setPhaseId(behaviorEvent.getPhaseId());
 				super.queueEvent(te);
 
 				return;
 			} else if ("rangechanged".equals(eventName)) {
 				// get start / end date
-				Calendar calendar = Calendar.getInstance(TimeZone.getTimeZone("UTC"));
-				TimeZone timeZone = ComponentUtils.resolveTimeZone(getTimeZone());
+				TimeZone targetTZ = ComponentUtils.resolveTimeZone(getTimeZone());
+				TimeZone browserTZ = ComponentUtils.resolveTimeZone(getBrowserTZ(), targetTZ);
 				TimelineRangeEvent te =
 				    new TimelineRangeEvent(this, behaviorEvent.getBehavior(),
-				                           DateUtils.toUtcDate(calendar, timeZone, params.get(clientId + "_startDate")),
-				                           DateUtils.toUtcDate(calendar, timeZone, params.get(clientId + "_endDate")));
+				                           DateUtils.toUtcDate(browserTZ, targetTZ, params.get(clientId + "_startDate")),
+				                           DateUtils.toUtcDate(browserTZ, targetTZ, params.get(clientId + "_endDate")));
 				te.setPhaseId(behaviorEvent.getPhaseId());
 				super.queueEvent(te);
 
@@ -543,6 +551,7 @@ public class Timeline extends UIComponentBase implements Widget, ClientBehaviorH
 		           .equals(context.getExternalContext().getRequestParameterMap().get(Constants.PARTIAL_SOURCE_PARAM));
 	}
 
+        @Override
 	public String resolveWidgetVar() {
 		final FacesContext context = FacesContext.getCurrentInstance();
 		final String userWidgetVar = (String) getAttributes().get(PropertyKeys.widgetVar.toString());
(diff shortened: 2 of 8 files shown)


ova2:
Sorry, but I could not understand what is the difference between the existing timeZone attribute and your browserTZ. The timeZone attribute already defines a user specific (local) time zone. It has the same meaning as your browserTZ. We already discussed this several times: dates on the server-side are in UTC. They are stored in DB and loaded from DB as UTC. It can only be UTC on the server-side. For displaying in browser (client-side) there is the timeZone attribute which is used for conversion to the user local time zone. Why do we need another attribute?

brooksie:
The (existing) Timezone parameter in the Timeline tag is the same as now - the timezone that you want to display the events in (target TZ)

BrowserTZ is the timezone that your browser is currently running in. 

When the JS create the timeline it always uses the TZ that your browser is running in. It assumes all the dates are provided in UTC, and automatically adds the offset before displaying.

When you want to display in a different TZ, you have to add the target TZ to UTC and subract the BrowserTZ, so that when the JS adds the BrowserTZ again, you get the right result. 

Please refer to http://forum.primefaces.org/viewtopic.php?f=14&t=30873&p=98658&hilit=offset#p98658 

And also to https://github.com/primefaces-extensions/core/pull/4
 Thanks.
    Steve.

ova2:
Hi Steve,

I thought JS Date can only understand UTC. AFAIK, a Date (JS or Java) can be only in UTC. But ok, I think I understand the issue. The problem is only, the Git says "We can’t automatically merge this pull request." This is because you used the 0.7.1 version. Can you please provide the same patch for the current 1.0.0 version? And please rename browerTZ to browserTimeZone (better attribute name).

I appreciate your effort and your help! Many thanks.

ova2:
I'm still not sure if we need browserTimeZone. I can imagine that users set their preferred time zone (somewhere in user settings), but who should define the browserTimeZone? In all web apps I've developed there was only one time zone (in user settings).

ova2:
It seems we really need an additional "browserTimeZone" attribute. Can you please provide a patch for the 1.0.0 version? I can test it asap then, for all use cases I have. Thanks.

brooksie:
Will work on it.

 
    Steve.

brooksie:
I've discovered new things about Git, updated the code (I hope) for V1.0 and sent in a pull request.

Suggest to test it around the start and end of daylight savings. If it's going to go wrong - it will be at these time.
http://www.timeanddate.com/worldclock/meeting.html is a good website for finding equivalent times in different timezones.

There was a bug reported in the forum recently about displaying the same date twice on the timeline for a brazilian timezone, again around daylight savings change. I haven't tested it, but I suspect this fix will help resolve that issue as well.

Happy testing!

     Steve.

ova2:
Hi, thanks, but why did you close  the pull request? :-) You should not close this issue, otherwise I can not merge the pull request. Can you open this issue again and send us a pull request? How to send a pull request is described here https://help.github.com/articles/using-pull-requests

One question yet. What is the fallback for the browserTimeZone when it is not set? The fallback should be taken then from the timeZone attribute in my opinion. What do you think? 

By the way, an automatic detection of the browser time zone would be great, but I read that it is not reliable.

brooksie:
Hi Oleg,

I thought there was a pull request here --> https://github.com/primefaces-extensions/core/pull/8 ?

I agree... the default for browserTimeZone is whatever "timeZone" is set to. That's how I coded it.
Fully agree that automatic detection of the browser's time zone would be superb.

 
    Steve.

ova2:
The pull request looks quite good. I tested it and the result was very good. It seems you have provided a good patch. I will do more tests (because I used the same logic at some other places) and commit then.

ova2:
Steve, I think something is still wrong. We said "if browserTimeZone is null (not set), it defaults to timeZone". That means browserTimeZone and timeZone are equal. Look your changes in DateUtils.java now:

```
// convert from local date to UTC
public static Date toUtcDate(TimeZone browserTZ, TimeZone targetTZ, Date localDate) {
    if (localDate == null) {
        return null;
    }

    long local = localDate.getTime();
    int targetOffsetFromUTC = targetTZ.getOffset(local);
    int browserOffsetFromUTC = browserTZ.getOffset(local);

    return new Date(local - targetOffsetFromUTC + browserOffsetFromUTC);
}

// convert from UTC to local date
public static long toLocalDate(TimeZone browserTZ, TimeZone targetTZ, Date utcDate) {
    long utc = utcDate.getTime();
    int targetOffsetFromUTC = targetTZ.getOffset(utc);
    int browserOffsetFromUTC = browserTZ.getOffset(utc);

    return utc + targetOffsetFromUTC - browserOffsetFromUTC;
}
```

If `targetTZ` is equal `browserTZ`, `toUtcDate(...)` returns `new Date(local)` and `toLocalDate(...)` returns `utc`. I believe this is not correct. What do you think? I think `browserOffsetFromUTC` should be 0 in this case. Can you confirm this?

I would say, if browserTimeZone is null (not set), it should stay null and `browserOffsetFromUTC` in the calculations above is 0 then.

Thanks. Oleg.

ova2:
Found some cool JS libs and would like to share the info with you.

Detection of browser's time zone in JavaScript is easy with this library https://bitbucket.org/pellepim/jstimezonedetect

Conversion from UTC to local time and back is easy with this library https://github.com/sproutsocial/walltime-js Example from this article http://engineering.sproutsocial.com/2013/02/javascript-time-zone-conversion-with-walltime/

```
var someUTCDate = new Date(),
chicagoWallTime = WallTime.UTCToWallTime(someUTCDate, "America/Chicago"),
backToUTCTIme = WallTime.WallTimeToUTC("America/Chicago", chicagoWallTime);
```

So, my idea is to combine these two libs. I think we don't even need the first lib, we only need the walltime lib and the conversion according to the provided "timeZone" attribute. So, we should think about this.

brooksie:
Yes,  if target TZ and browserTZ are the same, the net effect is zero and the timeline will display in local (browser) time.

In the existing code, the timezone is taken into account, a conversion is done, but the answer is wrong (your browserTZ offset gets added twice!). In the new code, the answer will be correct in local timezone, but the timezone conversion will have in effect been ignored, because the browserTZ was not specified. But without the browserTZ you can't do the correct conversion, so which wrong answer do you prefer??? If you want to replicate today's situation (and still get the wrong time conversion), the default for browserTZ should be UTC (i.e. zero, as you said).

I would do this by replacing any occurrence of:          

    TimeZone browserTZ = ComponentUtils.resolveTimeZone(timeline.getBrowserTimeZone(), targetTZ);
by                 

    TimeZone browserTZ = ComponentUtils.resolveTimeZone(timeline.getBrowserTimeZone(), 

        (timeline.getTimeZone()==null?targetTZ:TimeZone.getTimeZone("UTC")));

Note that if TimeZone is also null, you want TimeZone and browserTimeZone to have the same default, so that youdo get local timedisplayed.

My preference is to stick with the current default (browserTZ defaults to targetTZ). To me it's less wrong than the current situation... but there's a definite backwards compatibility issue here. As you said in an earlier note, if only there was a reliable way to detect the browserTZ, without requiring it as a specific input.

    Steve.

ova2:
Steve, I think there is a mistake. If target TZ and browserTZ are the same, the timeline will display UTC time. Look in the DateUtils.java. E.g. if timeZone = "Europe/Berlin" and we didn't specify browserTimeZone (that means it is "Europe/Berlin" too), the method toLocalDate(...) returns the UTC time and not my local time for Germany.

I added a time zone switcher to the showcase and a switch of time zone has no effect with your patch. Timeline's start, end, min, max dates shows always the same time for all time zones. They are displayed in UTC. Also the red line (current time) always shows me the UTC time and not my local time. Something is wrong here definitely.

brooksie:
I'm concerned that you say the dates are displayed in UTC. That's not my experience. Perhaps you can send me a screen shot and the code so I can have a look. Unfortunately I'm away on business all this week. I can view the code but won't be able to test or change anything until after I get back.

An alternative is to use the JS you found as a much better default - but I have no idea how to go about integrating that. You need the answer from the JS at the point the bean runs to generate the code. Do you know how to?
Thanks.

ova2:
I can post screens later. But again, how should this work if `toLocalDate(...)` returns the UTC due to the same offset for target TZ and browser TZ? I mean in the return statement we have `utc + targetOffsetFromUTC - browserOffsetFromUTC` what is equal `utc`.

brooksie:
I still think it's OK. Everything is done in UTC... dates are always in UTC... and the timeline JS converts UTC to browser time on display, and back again on events. The bean doesn't convert... the timeline JS does.  That's why I'm concerned when you say the display is in UTC. If that really is the case then the theory above is incorrect :-( . But being a good IT professional, I don't believe it until I see it for myself - I'm sure you understand ;-)
 
I think the misconception is thatlocal date isn't really local date.... it is adjusted UTC date - you're adjusting UTC to reflect the other (=target) timezone. What we adjust by is the timezone offset between the targetTZ and the browserTZ for that UTC date. But we can't do it directly, so we do it by the offset between target TZ and UTC and the browserTZ and UTC. Why can't we just adjust by the targetTZ offset? ... because you're not living in UTC, your living in local time and it's the difference between target and browser that you need to know. And the timeline JS will always adjust by browserTZ offset. So we must take that into account when calculating the difference to simulate a display in the targetTZ. We might think we're looking at a timeline that shows Chicago time, but we're still really in Berlin and the browser is still running on Berlin time.
 
The way it's written right now... if targetTZ and browserTZ are the same, UTC and local date will be the same, the timeline JS will add the browser offset automatically, the event will display in the correct local time. 
And if targetTZ is set but browserTZ isn't, then browserTZ defaults to targetTZ, and the event displays in local time... in effect completely ignoring the targetTZ (TimeZone attribute) setting. Per my previous note, is this more wrong than the current method of calculating an incorrect offset? Without knowing the browserTZ, you can't get it right.
 
If you're still not convinced... send me the evidence. 
 
Using the JS library to provide a default browserTZ will help to resolve the problem. 
Alternatively changing the timeline JS to display in UTC rather than local time would also resolve the issue (although introducing other anomalies around the start/end of summertime). 
    
    Steve.

ova2:
There were some discussions in the forum for the native JS widget. E.g. this one https://github.com/almende/chap-links-library/issues/31 Timeline widget is not time zone aware currently. Doing the conversion on the server-side is ok for me, at least for now. As I said, I have added a time zone switcher to this example http://fractalsoft.net/primeext-showcase-mojarra/sections/timeline/dnd.jsf and today it started to work suddenly :-). You can try the same

XHTML

```
<pe:timeline id="timeline" value="#{dndTimelineController.model}" var="event"
             editable="true" eventMargin="10" eventMarginAxis="0" minHeight="250"
             start="#{dndTimelineController.start}" end="#{dndTimelineController.end}"
             timeZone="#{dndTimelineController.localTimeZone}"
             style="margin-left:135px;" snapEvents="false" showNavigation="true"
             dropActiveStyleClass="ui-state-highlight" dropHoverStyleClass="ui-state-hover">
    <p:ajax event="drop" listener="#{dndTimelineController.onDrop}" global="false" update="eventsList"/>

    <h:panelGrid columns="1">
        <h:outputText value="#{event.name}"/>
        <h:outputText value="#{event.start}">
            <f:convertDateTime dateStyle="short" type="both" timeZone="#{dndTimelineController.localTimeZone}"/>
        </h:outputText>
        <h:outputText value="#{event.end}">
            <f:convertDateTime dateStyle="short" type="both" timeZone="#{dndTimelineController.localTimeZone}"/>
        </h:outputText>
    </h:panelGrid>
</pe:timeline>

<!-- TIME ZONE SWITCHER -->
<h:panelGrid columns="2" style="margin-top:15px;">
    <h:outputText value="Choose your local time zone"/>
    <p:selectOneMenu value="#{dndTimelineController.localTimeZone}" style="margin-left:7px;">
        <f:selectItems value="#{dndTimelineController.timeZones}"/>
        <p:ajax update="timeline"/>
    </p:selectOneMenu>
</h:panelGrid>
```

Java

```
@ManagedBean
@ViewScoped
public class DndTimelineController implements Serializable {

    private TimelineModel model;
    private Date start;
    private Date end;
    private String localTimeZone = "Europe/Berlin"; // initial time zone
    private List<SelectItem> timeZones;

    private List<Event> events = new ArrayList<Event>();

    @PostConstruct
    protected void initialize() {
        // set initial start / end dates for the axis of the timeline
        Calendar cal = Calendar.getInstance(TimeZone.getTimeZone("UTC"));
        Date now = new Date();

        cal.setTimeInMillis(now.getTime() - 4 * 60 * 60 * 1000);
        start = cal.getTime();

        cal.setTimeInMillis(now.getTime() + 8 * 60 * 60 * 1000);
        end = cal.getTime();

        // create timeline model
        model = new TimelineModel();

        // create available events for drag-&-drop
        for (int i = 1; i <= 13; i++) {
            events.add(new Event("Event " + i));
        }

        // get all unique time zones (code of the TimeZoneUtil is not relevant here)
        timeZones = new ArrayList<SelectItem>();
        for (TimeZone timeZone : TimeZoneUtil.getTimeZones()) {
            timeZones.add(new SelectItem(timeZone.getID(), TimeZoneUtil.getName(timeZone)));
        }
    }

    public void onDrop(TimelineDragDropEvent e) {
        // get dragged model object (event class) if draggable item is within a data iteration component,
        // update event's start and end dates.
        Event dndEvent = (Event) e.getData();
        dndEvent.setStart(e.getStartDate());
        dndEvent.setEnd(e.getEndDate());

        // create a timeline event (not editable)
        TimelineEvent event = new TimelineEvent(dndEvent, e.getStartDate(), e.getEndDate(), false, e.getGroup());

        // add a new event
        TimelineUpdater timelineUpdater = TimelineUpdater.getCurrentInstance(":mainForm:timeline");
        model.add(event, timelineUpdater);

        // remove from the list of all events
        events.remove(dndEvent);

        FacesMessage msg = new FacesMessage(FacesMessage.SEVERITY_INFO, "The " + dndEvent.getName() + " was added", null);
        FacesContext.getCurrentInstance().addMessage(null, msg);
    }

    public TimelineModel getModel() {
        return model;
    }

    public void setLocalTimeZone(String localTimeZone) {
        this.localTimeZone = localTimeZone;
    }

    public String getLocalTimeZone() {
        return localTimeZone;
    }

    public List<SelectItem> getTimeZones() {
        return timeZones;
    }

    public List<Event> getEvents() {
        return events;
    }

    public Date getStart() {
        return start;
    }

    public Date getEnd() {
        return end;
    }
}

public class TimeZoneUtil {

    private static final String TIMEZONE_ID_PREFIXES = "^(Africa|America|Asia|Atlantic|Australia|Europe|Indian|Pacific)/.*";

    private static List<TimeZone> TIME_ZONES;

    public static List<TimeZone> getTimeZones() {
        if (TIME_ZONES == null) {
            TIME_ZONES = new ArrayList<TimeZone>();

            final String[] timeZoneIds = TimeZone.getAvailableIDs();

            for (final String id : timeZoneIds) {
                if (id.matches(TIMEZONE_ID_PREFIXES)) {
                    TIME_ZONES.add(TimeZone.getTimeZone(id));
                }
            }

            Collections.sort(TIME_ZONES, new Comparator<TimeZone>() {

                    public int compare(final TimeZone t1, final TimeZone t2) {
                        return t1.getID().compareTo(t2.getID());
                    }
                });
        }

        return TIME_ZONES;
    }

    public static String getName(TimeZone timeZone) {
        return timeZone.getID().replaceAll("_", " ") + " - " + timeZone.getDisplayName();
    }
}
```

The switcher switches the start / end dates very well. But the current time (red line) is the same. I have to call `setCurrentTime(time)` manually (see docu here http://almende.github.io/chap-links-library/js/timeline/doc/) to show the current time in the user defined time zone and not the browser's one. I will do this and commit then your patch. Another problem is that we can not use JSF date / time converter, e.g.

```
<h:outputText value="#{event.start}">
     <f:convertDateTime dateStyle="short" type="both" timeZone="#{dndTimelineController.localTimeZone}"/>
</h:outputText>
```

because it doesn't consider browser time zone. But ok. Thanks for the great effort!

ova2:
Found another problem. Timeline JS seems doesn't convert local time to UTC back on events. The example with drag-&-drop shows this. I receive the local time on the server-side when a new event is added.

ova2:
Strange, but now it doesn't work again with the time zone switcher. It always shows the same start / end times. Can you try the code above please if you have time?

Edit: It was working with my old code. I tested with my old code and forget to recompile. It doesn't work with your patch. Simple try the switcher above. By the way, Timeline JS doesn't do itself any conversions internally.

brooksie:
Hi again. 

I looked at this and I am having similar flaky issues, that I can't yet tie down. My cut-down version seems to be sort-of OK if I've run another PF or PFE transaction first, then used the switcher functionality. But GlassFish reports render errors if this is the first txn that is used on loading in the browser after the index page (which has no PF or PFE in it).

Refresh also only works if <p:ajax update="@form"> is used, not update="timeline". ?? is there a hidden block somewhere?

The bad news is, the same symptions happen when I'm using vanilla PFE 1.0, so the problem doesn't appear to be related to my timeZone changes.

The red line issue is a "feature", I'm afraid. PFE adjusts the UTC dates to pretend we're in another timezone - the red line is put in by the JS, and it is not aware that we've done that - it's only aware of the browser's TZ. So, bottom line, the red line will always be wrong if you're not displaying the timeline based on your local browser
 TZ, so don't use it.

The JSF Date Converter issue: I think the issue is related to when the TimeZone expression is evaluated - only once on first page load. Hence why it never changes. The event display can be made to change if you use the TimelineUpdater functionality.
 
    Steve.

ova2:
Red line is a special case, I will correct this, but the boundaries are set by min. / max. values and they are converted by your logic. And they always display the same times for all time zones. If I remove your logic (back to the current one), boundaries are ok.

JSF Date Converter is used when a new event is added (by drag-&-drop or double click). Not only once on the first page load. You can debug if you want. It is always called.

brooksie:
I can't get drag and drop to work at all on my system! 
DnD works OK if I execute the demo from your server. On my system (=PC), the drag is OK and the timeline highlights. But the drop event never gets fired in my bean. It also seems to re-execute the @PostConstruct initialisation() function every time. Not sure why - the bean is @ViewScoped. This is true if I revert to vanilla PFE 1.0 as well. It may well be unrelated to timeline and true of any DnD funtionality... I haven't check this yet.
I'm using Firefox 24.0, Netbeans 7.3, Glassfish 3.1.2, Mojarra 2.1.6, PF4.0 and PFE1.0 - any suggestions?

Meanwhile, I notice your don't have browser timezone set. If you set the browser-timezone attribute, does it work better? When the browser Timezone and the target timezone are the same, or browser-Timezone is missing, the timezone attribute is effectively ignored and the timeline always displays in the local timezone... hence one reason why nothing might appear changed on the display. 

    Steve.

ova2:
Try with Mojarra 2.1.24 or higher please. It could be a bug in Mojarra or GlassFish 3.1.2. It works for me local with JBoss 7 and Jetty Maven plugin (mvn jetty:run). No issues.

"When the browser Timezone and the target timezone are the same, or browser-Timezone is missing, the timezone attribute is effectively ignored and the timeline always displays in the local timezone... hence one reason why nothing might appear changed on the display."

Yes, I tried to say you the same :-). This is not an expected behaviour in my opinion.

brooksie:
Thanks for the hint on Mojarra. Upgrading to 2.1.26 resolved several issues including calling the @PostConstruct init() function multiple times. It also resolved not being able to invoke the page immediately on entry to the application - the flakiness has gone away. 

What it did not resolve, though, was invoking the onDrop event. What has resolved that was adding a "widgetVar" attribute to the timeline tag. In fact the absence of that attribute was stopping all timeline events from firing.  The same fix also resolved the need to use @form in the ajax statement on updating the switcher. It works OK now with update="timeline". I don't know why this is the case... some problem before reaching Timeline.queueEvent() because that wasn't being executed. Nothing in the server log. However it all works now. 

Back to the question in hand. What you don't like is the default for browserTime. Ok. I've change it to default to the server TZ, same as for timeZone. If your browser is running on the same TZ as the server, it will be OK, but if your in a different TZ, bad luck, the display will be wrong (unless you've explicitly specified the attribute). I _think_ I've added a new commit to the pull request. Try it and see.

    Steve.

ova2:
Hi Steve,

Thanks for the reporting the issue with widgetVar. I will look into this deeper. I see your commit. Will test it as soon as possible.

But again, the same question regarding conversion in DateUtils.java. If target TZ and browser TZ are defined and the same (e.g. Germany), the local TZ will be UTC according to the code. Is it correct? I believe not. The same for the conversion UTC --> local TZ.

brooksie:
Yes it is correct... and the time will be correctly displayed in LOCAL time. 
Eg. 13:00 Germany today (daylight savings has ended) is 12:00 UTC. So UTC date is 12:00.

Conversion to local time is 0, as you say, meaning local date is also 12:00. 

And 12:00 is intepreted as UTC by the JS and displayed in the browser as 13:00 German time (if your browser is running on German time).

The key is that local date isn't really a local date. It's still a UTC date, but offset by the time difference between Target TZ and Browser TZ.
 

    Steve.

ova2:
Hi Steve,

Good news. All time zone issues are fixed now! I committed all changes. Thanks for the patch.

I could not reproduce the issue with the widgetVar. It works fine without widgetVar. But I fixed the issue with displaying the current time (red line). Check the SNAPSHOT showcase if you want. I have added a second drop-down for choosing browser's time zone.

PR #8, opened 2013-10-10
Further updates for v1.0

brooksie:
Please check the merge carefully - particularly the taglib changes - I'm not confident with Git, and hope I've got this right without corrupting anything :-)
I changed the new attribute to be "browserTimeZone" but continued to use "TZ" in the java code.
From a user's perspective, if you want events displayed in local (browser) time, leave both TimeZone and BrowserTimeZone empty, OR fill both with the same timeZone. If you want to display events using a different (target) timezone than the browser uses, both TimeZone and BrowserTimeZone attributes must be supplied for the time conversion to be correct. If you only supply the TimeZone attribute (any pre-existing code), the timeline display will be local time, because browserTimeZone defaults to the same as TimeZone. This is different from what happens today.


ova2:
Merged as discussed here https://github.com/primefaces-extensions/core/pull/5

PR #9, opened 2013-12-04
Added timestamp to error output

jherkel:
I would like to add a timestamp to the error output. From time to time we need to find an exception also in logs and with the timestamp it is much easier, especially when we can see the same exception more times.

### DIFF
--- a/src/main/java/org/primefaces/extensions/component/ajaxerrorhandler/AjaxExceptionHandler.java
+++ b/src/main/java/org/primefaces/extensions/component/ajaxerrorhandler/AjaxExceptionHandler.java
@@ -38,6 +38,8 @@ import java.util.Map;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import org.apache.commons.lang3.exception.ExceptionUtils;
+import java.text.SimpleDateFormat;
+import java.util.Date;
 
 /**
  * {@link ExceptionHandlerWrapper} which writes a custom XML response for the {@link AjaxErrorHandler} component.
@@ -49,6 +51,7 @@ import org.apache.commons.lang3.exception.ExceptionUtils;
 public class AjaxExceptionHandler extends ExceptionHandlerWrapper {
 
 	private static final Logger LOGGER = Logger.getLogger(AjaxExceptionHandler.class.getCanonicalName());
+	private final SimpleDateFormat dateFormatter = new SimpleDateFormat("yyyy-MM-dd HH:mm:ss");
 
 	private ExceptionHandler wrapped = null;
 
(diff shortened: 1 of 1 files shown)


ova2:
Pavol. Can we merge this patch?

slany:
Hi, I cannot recommend this patch, because:
  1) The format of timestamp output has to be configurable 
  2) SimpleDateFormater is not synchronized (it is not thread safe) => this patch is not OK.

Oleg, - what is your opinion?

jherkel:
Hi,

Ad 1) I don't agree with your opinion. Why do you want to customize output format? Maybe there can be info about milliseconds, but it is information for developers not for the end user. In an ideal world the end user shouldn't  ever see it. And of course it is not a log file, when I need to check thousands of lines, I would probably customize my output format. 
Ad 2) Good point. My mistake, I can replace it with FastDateFormat

Jakub

ova2:
This is the discussion https://github.com/primefaces-extensions/primefaces-extensions.github.com/issues/197

Because Pavol is the developer of this component, he should decide if we will accept or reject this pull request. So, what is your decision Pavol?

fenoloco:
I use to generate an unique id to match screen error message with log messages.
Maybe that could be a solution.

slany:
Hi  jherkel,
About first point. Every public library has to work for a lot of people with difference languages and UI habits. If there is DATE-TIME output, we have to allow add formatting this value (default value can be ISO format). We cannot ignore this option and prepare something what we has to update later.

jherkel:
I can see a lot of similarity with stacktrace in this case. Again - it is NOT information for the end user. The end user, of course, knows his time and probably he has time in a localized format. So your whole idea about different "languages and UI habits" is wrong. For example, you set up time format output and in CZ it will produce something like "Pondeli 4 rijen 2013" and in EN "Monday 4 October 2013". The end user will send data from this to a developer and he will be "very happy" about your idea "Every public library ...".  Linux kernel panic isn't localized, BSOD isn't localized, so could you give me an example what you want to have localized? 

Anyway, if you believe that localized time is a good idea and it is the only problem with this patch, I can prepare the second version.

ova2:
So, Pavol, should be the second version prepared?

slany:
Hi JHerkel - I understand to your opinion, but not agree ... If LOG4J has option for timestamp formating, we will have the same option. You have to understand that this is only option, which can be ignored, because default format for timestamp will be the same as you suggested.

This patch will be closed without merging. Merge pull request will be done by command line with fixing thread safe bug and adding option for timestamp formatting.

slany:
This features is pushed to the repository.

PR #12, opened 2014-09-21
Added missing height property to knob component

frederikb:
The knob component is missing a height property to match the data-height property of the Knob jQuery plugin. This property isn't explicitly documented on the documentation site, but is used as part of the "big" example. Setting the height is necessary for large knobs or else the result will appear cut off.

`<pe:knob/>` using a width setting of 500 and no height specified
knob_cutoff


ova2:
Look good. Thanks.

PR #13, opened 2014-11-25
Update primefaces-extensions.taglib.xml

realdognose:
readonly is inherited from HtmlInputText, therefore not camel cased. (other readOnlys might require review as well)


tandraschko:
Thanks for the pull request - they are allways welcome :)
But i already fixed it!

PR #14, opened 2014-12-11
improve readability by using more specific interfaces instead of generic Map

nemecec:
Simple change, doesn't change much, just readability. I actually wanted to use ConcurrentMap.computeIfAbsent() method but unfortunately, this is only available since Java 1.8 but Primefaces needs to support also 1.5.


tandraschko:
Hi Neeme,

thanks for you pull request. It looks fine.
I will answer try the new plugin next week + answer your mails. This  week is a little big stressfull...

PR #17, opened 2015-05-20
Don't assume 1 sec passed in doStep, calculate it

unclesnottie:
Javascript `setInterval()` notoriously unreliable at executing at a precise time (see: http://ejohn.org/blog/how-javascript-timers-work/).  `doStep()` is assuming that since the interval is 1000ms, the countdown timer has changed by 1s.  That isn't always the case.  For example, on Mac, running a timer in your browser then locking the screen can cause the timer to fall behind.

This fix adds a new function `currentTimeInSecs()`.  It is used in `start()` to store the current time as `prevTime`.  Then in `doStep()` the time passed is calculated by subtracting `prevTime` from the current time in seconds and `prevTime` gets updated.  This allows `doStep` to be missed or called later than expected without the `currentTimeout` being wrong.

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces-extensions/timer/timer.js
+++ b/src/main/resources/META-INF/resources/primefaces-extensions/timer/timer.js
@@ -19,6 +19,9 @@ PrimeFacesExt.widget.Timer = PrimeFaces.widget.BaseWidget.extend({
         this.print();
 
     },
+    currentTimeInSecs: function() {
+      return parseInt((new Date()).getTime() / 1000, 10);
+    },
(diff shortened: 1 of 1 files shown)


ova2:
Francesco, is it okey for your? Can we approve the pull request?

francesco-strazzullo:
Yeah no problem with me! This week I can't test it, but you can already merge in my opinion!

ova2:
Merged. Thanks for the fix.

unclesnottie:
Awesome.  So glad I could help!

PR #18, opened 2015-05-21
KeyFilter improvement: InputRegEx to have regEx on the whole text

Julienng:
I create a new Parameter to add regEx like `/^[0-9]{0,3}((\\.|,)[0-9]{0,6}|)$/`.
- Parameter inputRegEx.
- Function inputFilter based on event Jquery 'input'.

### DIFF
--- a/src/main/java/org/primefaces/extensions/component/keyfilter/KeyFilter.java
+++ b/src/main/java/org/primefaces/extensions/component/keyfilter/KeyFilter.java
@@ -62,7 +62,7 @@ public class KeyFilter extends UIComponentBase implements Widget {
 		widgetVar,
 		forValue("for"),
 		regEx,
-		regExAll,
+		inputRegEx,
 		mask,
 		testFunction,
 		preventPaste;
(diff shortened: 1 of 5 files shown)


ova2:
Thomas, is it ok for you? Can we approve the pull request?

tandraschko:
Thanks for the pull request.
Would you please move the logic from $.fn.inputFilter to the 1-keyfilter.js file?
Thats because 0-jquery.keyfilter.js is the jquery plugin and we possible overwrite your changes if we update it in the future.

Julienng:
What did you mean ? Like move my logic into a function in `keyFilter` and pass it like a callback into `inputFilter` ?

tandraschko:
i would remove the jquery function and directly apply the .on the inputfield.

input.inputFilter(this.cfg.inputRegEx);

->

input.on('input', function(e) {.....});

Julienng:
Ok, I would do it.

Julienng:
I doesn't have enough time last week. It's uglier than before but work as expected.

tandraschko:
Looks good from my perspective. i will merge it :)

Julienng:
Good news. :smile:  Just in case: every regEx into inputRegEx have to start with `^` and finish with `$` to match the complete string.

PR #20, opened 2015-11-26
Bug Fix: exporter exports only exportable columns

yelhouti:
In the group columns exporter used to export columns with exportable false, this intended to fix that problem.


sudheerj:
Did you create an issue for this PR?

yelhouti:
I don't remember doing so, thank you for the merge.

PR #21, opened 2016-01-26
Added attribute 'zoom' for documentViewer

olpf:
Added the attribute 'zoom' for documentViewer. Attributes value is added in the iframes url like the 'page' attribute. Possible values for the attribute can be taken from the pdf.js documentation. E. g. numerical values for the zoom like '75' or 'page-width', 'page-height', 'page-fit' or 'auto'. If the 'zoom' attribute is not specified the documentViewer behaviour does not change.


ova2:
Thanks.

PR #22, opened 2016-02-10
[Perf] No need to keep looping once caseToRender has been set

Rapster:
Fix for #369


tandraschko:
the PR looks fine but please create a issue for tracking.

Rapster:
Don't have access to the "Issues" menu item :/

tandraschko:
It's here: https://github.com/primefaces-extensions/primefaces-extensions.github.com/issues

Rapster:
Done: #369

Rapster:
By the way, is there any performance added-value by using switch rather than consecutive `ui:fragment`

tandraschko:
Don't think so.
Applied - thanks for the PR!

PR #25, opened 2016-05-06
DynaFormRow should support nested DynaFormModel

Rapster:
Here is a preview of what I've done so far. Thanks for your feedbacks.


ova2:
It looks ok. It would be good to mention the nested models in the showcase too. But ok.

Rapster:
I'll do it ;) Thanks @ova2

ova2:
Super! Thanks. My name is Oleg by the way.

PR #27, opened 2016-05-08
JavascriptBehavior extends AbstractBehavior

CSchulz:
Fixed primefaces-extensions/primefaces-extensions.github.com#377

I was not able to build this locally cause I am missing the primefaces 6 snapshot artifact.


ova2:
Thomas, can you please check it?

PR #28, opened 2016-05-12
Support of DynaForm columnClasses

Rapster:
And some renaming rearding DynaForm nested model


src/main/java/org/primefaces/extensions/component/dynaform/DynaFormRenderer.java
```diff
@@ -254,37 +261,37 @@ protected void encodeBody(FacesContext fc, DynaForm dynaForm, List<DynaFormRow>
 					}
 
 					writer.endElement("label");
-				}  else if(element instanceof DynaFormNestedModel){
-					DynaFormNestedModel nestedModel = (DynaFormNestedModel) element;
-					
+				} else if (element instanceof DynaFormModelElement) {
```
ova2, line 264: Can we check for DynaFormControl first and then DynaFormModelElement? In 99% DynaFormControl will be used, so that the else if with DynaFormControl should be better placed before the else if with DynaFormModelElement. Ok?

ova2:
I made a comment regarding else if check (small perf. optimization). Can you look this please? Thanks.

I also hope that all changes will not break backwards compatibility because my client uses this component intensive.

PR #29, opened 2016-05-13
Evaluate DynaFormModelElement as the last eligible case

Rapster:


ova2:
Very good. Thanks for your effort!

ova2:
By the way, Sebastien, should I add you as committer to PF Extensions? We have some inactiv members. I would remove inactive members and add you as committer. What do you think?

Rapster:
I really appreciate that Oleg, indeed I'm working on a big project using Primefaces and lately, DynaForm saved us a lot of trouble from memory issues: this is why I created a few ticket. I hope I could do more in the future as I really like to involve myself in Primefaces. As I'm not a JSF expert, I still have to understand some basics and other stuff, I think it is more wise that I continue to make pull request and someone review the code, at some point, I hope I'll have enough skills to become a committer to PF Extensions.

ova2:
Ok, all right.

By the way, I have planned to add this feature to DynaForm (responsive layout with DIVs instead table) https://github.com/primefaces-extensions/primefaces-extensions.github.com/issues/321 but didn't have interest then because changed my job and the new client doesn't use DynaForm :-) If you want, you can try to do this. Just FYI.

PR #30, opened 2016-05-13
Update gravatar base url to use https request

arthurgregorio:
PR to fix the warnings show on #372


arthurgregorio:
I made a new commit (a2954cd) with a configurable way to made the request secure

PR #31, opened 2016-06-09
Fix #384

Rapster:
From https://github.com/primefaces/primefaces/issues/998


src/main/java/org/primefaces/extensions/component/base/AbstractDynamicData.java
```diff
@@ -327,6 +313,40 @@ public void processUpdates(FacesContext context) {
...
 
+    protected void preDecode(FacesContext context) {
+        Map<String, SavedEditableValueState> saved = (Map<String, SavedEditableValueState>) getStateHelper().get(PropertyKeys.saved);
+        if (null == saved || !keepSaved(context)) {
+            getStateHelper().remove(PropertyKeys.saved);
+        }
+    }
+
+    private boolean keepSaved(FacesContext context) {
+        return (contextHasErrorMessages(context) || isNestedWithinIterator());
```
Rapster, line 324: The big change is right here, calling `isNestedWithinIterator()`

PR #32, opened 2016-07-05
The web browser language identification codes and Java locale values aren't the same in most cases.

JuanMiguelBG:
Example:
- Spanish Web browser language identification codes: es or es-ES
- Spanish Java locale : es or es_ES.

When I translate from web code to Java locale, we have to change the '-' with '_' character.
When I translate from Java locale to web code, we have to change the '_' with '-' character.


ova2:
Merged

PR #33, opened 2016-08-12
Excel Exporter wuth column group export all the colum label without check if is rendered or exportable.

psini:
Excel Exporter wuth column group export all the colum label without check if is rendered or exportable.


ova2:
Merged

PR #36, opened 2016-10-24
Fix Issue #415 - set readOnly attribute to the config file after all …

jxmai:
…resources were loaded

For: https://github.com/primefaces-extensions/primefaces-extensions.github.com/issues/415


melloware:
Nice work just affect the instance.

PR #40, opened 2016-12-09
Update column type to Object to allow specifying data type and other …

masergy-mobile:
…options.  See https://developers.google.com/chart/interactive/docs/datatables_dataviews

This allows creating tooltips by creating a column with a tooltip role as described at https://developers.google.com/chart/interactive/docs/customizing_tooltip_content


melloware (approved):
Nice work.  Thanks for the PR.

PR #44, opened 2017-09-11
documentviewer: hashParams is undefined

KarlKnapp:


melloware:
How are you able to produce this error?  I would like to reproduce is before merging?   I test using the showcase and right now the showcase is working on the latest snapshot?

jxmai:
@KarlKnapp Could you please raise an issue in the https://github.com/primefaces-extensions/primefaces-extensions.github.com/issues repo with a sample for tracking purposes?

KarlKnapp:
The issue in the latest git version of pdf.viewer.js should be obvious because the variable 'hashParams' is used before it is defined.
@melloware The showcase does in deed not have this issue because the ' if ('locale' in hashParams)'-section is called after the initialization of hashParams. (pdf.viewer.js.jsf starting on line 20520 )
Which is the latest version? git-version or showcase-version?

melloware:
You are right it was related to this change:  https://github.com/primefaces-extensions/primefaces-extensions.github.com/issues/482  

I had fixed that locally and somehow accidentally reverted my change for locale before checking in.  Thanks for the fix!

melloware:
I just cut an Extensions 6.1.5 Elite release including this fix if you need it.  Read here:  https://forum.primefaces.org/viewtopic.php?f=14&t=52458

PR #45, opened 2017-10-12
Fix typo in widget.js

jxmai:
`an error` not `a error`


melloware:
yeah this should be "CKEditor threw an error while destroying the old instance:"

jxmai:
The reason I closed this is because there are multiple error places so I don't want to include multiple commits just to fix typo... Like there are two places in ckeditor.js  showing  `a error`

melloware:
I will fix!

jxmai:
Thanks :+1:

melloware:
https://github.com/primefaces-extensions/core/commit/df049387e32bd8f9fbee3cd3afea6763fd878cd2

PR #53, opened 2018-08-30
Fix #591: Sheet: Column filter focus problem

### LINKED ISSUE #591: Showcase: Add flags
The input phone component contains flag images which can be stand alone as well. We could add a page to the showcase to demo this.

rbetjes:
Proposed fix for "#591: Sheet: Column filter focus problem".
I'm not a fan of the used solution as setTimeout is in my opinion almost always a work-around instead of a fix.
However, I've done some testing and everything seems to work correctly with this in place.


melloware (approved):

melloware:
Thanks for investigating and the fix.  Yeah if you look on line 309 in the 1-sheet.js you will see we had to do something similar.  I think the timeout might be the only way to do this!

PR #54, opened 2018-09-03
New Component: orgchart

jxmai:
https://github.com/primefaces-extensions/primefaces-extensions.github.com/issues/480


src/main/java/org/primefaces/extensions/component/orgchart/DefaultOrgChartNode.java
```diff
@@ -0,0 +1,144 @@
...
+ * <code>orgchart</code> component.
+ *
+ * @author @jxmai / last modified by $Author$
+ * @version $Revision$
+ * @since 6.3
+ */
+public class DefaultOrgChartNode implements OrgChartNode {
+
+    private String id;
+
```
melloware, line 31: Should we implement equals(), hashCode() and toString() just in case someone wants to add these to a hashMap or set to track them or check two nodes equality?

melloware:
Nice work on this @jxmai.

src/main/java/org/primefaces/extensions/component/orgchart/DefaultOrgChartNode.java
Rapster: I think we could yes :wink:

melloware:
Also I know you are still working on this but adding Showcase examples would be the final thing once you are done all the work on the core component.

jxmai:
showcase sample: https://github.com/primefaces-extensions/showcase/pull/10 ( a simple sample for the time being)

melloware:
Nice!  Also can you add a line to this file...  https://github.com/primefaces-extensions/core/blob/master/src/main/resources/META-INF/NOTICE.txt

jxmai:
Thanks for the notice. https://github.com/primefaces-extensions/core/pull/54/commits/56f720aad9e55e21c6957d9a3a41a8c3f8b471ab

PR #56, opened 2018-10-11
Sheet: Adding filterMatchMode to pe:sheetcolumn

rbetjes:
Just a spontaneous enhancement, no official issue created for this.


melloware (changes requested):
Nice work.  You missed one file you need to add the new propery to src\main\resources\META-INF\primefaces-extensions.taglib.xml in the SheetColumn object.  Its down near the bottom of the file.

PR #64, opened 2019-02-26
Fix #650: added InputPhone

jepsar:


melloware (changes requested):
Don't forget you need to add it to the taglib.xml file too and all its properties.

src/main/java/org/primefaces/extensions/component/inputphone/Country.java
```diff
@@ -0,0 +1,46 @@
...
+ * limitations under the License.
+ */
+package org.primefaces.extensions.component.inputphone;
+
+/**
+ *
+ * @author Jasper de Vries &lt;jepsar@gmail.com&gt;
+ */
+public class Country {
+
```
melloware: Best to make this implement Serializable and add an equals(), toString(), and hashCode(). Whenevre you have these POJO's you never know how someone might use them or put them in SessionScoped beans etc.

melloware (changes requested):

jepsar:
> Don't forget you need to add it to the taglib.xml file too and all its properties.

Sorry, I'm not sure what you mean. I modified primefaces-extensions.taglib.xml. Did I miss something?

tandraschko (commented):
In pf core we would put this class into the model package

jepsar:
> In pf core we would put this class into the model package

I wasn't sure about this.. I'll move it.

melloware:
@jepsar Nice work.  For some reason the taglib was now showing up in my Diff but it is now!  Thanks.  I will merge tonight and send you a note when i have it all working with minifying everything correctly.

melloware (approved):

PR #65, opened 2019-03-01
Fix primefaces-extensions/primefaces-extensions.github.com#657

jepsar:


src/main/java/org/primefaces/extensions/component/inputphone/InputPhoneRenderer.java
```diff
@@ -150,6 +148,12 @@ protected void encodeScript(FacesContext context, InputPhone inputPhone) throws
         if (StringUtils.isNotEmpty(inputPhone.getInitialCountry())) {
             wb.attr("initialCountry", inputPhone.getInitialCountry());
         }
+        if ("auto".equals(inputPhone.getInitialCountry())) {
+            if (inputPhone.getGeoIpLookup() == null) {
+                throw new IllegalArgumentException("geoIpLookup is required with initialCountry 'auto'");
+            }
```
melloware, line 155: One minor change here.  Change to throw FacesException as that is typically what i have seen thrown from inside components.  Second lets change the message slightly to "InputPhone geoIpLookup property is required when initialCountry is 'auto'."

melloware (changes requested):

src/main/java/org/primefaces/extensions/component/inputphone/InputPhoneRenderer.java
jepsar: Done.

PR #66, opened 2019-03-06
Fix #659

### LINKED ISSUE #659: New Helper: Mongo DB Lazy Datatable implementation
Thanks to Newk we will be adding a Mongo DB based `LazyDatamodel` similar to the `JPALazyDataModel` added in PF 11.

jepsar:
Fix #659


tandraschko:
Please also add "//@Override" to destroy/refresh and move it below init.
Otherwise it looks like a public widget method.

melloware:
@tandraschko I just added the // @override thanks for noticing.  Checked in.

PR #68, opened 2019-03-13
Fix #663

jepsar:
Fix https://github.com/primefaces-extensions/primefaces-extensions.github.com/issues/663


pom.xml
```diff
@@ -437,6 +437,13 @@
 			<artifactId>poi-ooxml</artifactId>
 			<scope>provided</scope>
 		</dependency>
+        <!-- InputPhone dependencies -->
+        <dependency>
+            <groupId>com.googlecode.libphonenumber</groupId>
+            <artifactId>libphonenumber</artifactId>
+            <version>8.10.7</version>
+            <scope>provided</scope>
```
jepsar: So, I added this dependency as provided. I'm not sure if we can just add it (Apache 2, 340kB)

src/main/java/org/primefaces/extensions/validate/PhoneValidator.java
```diff
@@ -0,0 +1,59 @@
...
+import javax.faces.validator.ValidatorException;
+import org.primefaces.extensions.component.inputphone.InputPhone;
+
+/**
+ * Validator used with {@link InputPhone}.
+ *
+ * @author Jasper de Vries &lt;jepsar@gmail.com&gt;
+ */
+@FacesValidator("org.primefaces.extensions.validate.PhoneValidator")
+public class PhoneValidator implements Validator {
```
jepsar, line 34: I'm not sure if we want to validate by default, so I opted to create a validator class which can be used optionally. We could move the code to the renderer and validate based on an attribute flag.

src/main/java/org/primefaces/extensions/validate/PhoneValidator.java
```diff
@@ -0,0 +1,59 @@
...
+import javax.faces.validator.Validator;
+import javax.faces.validator.ValidatorException;
+import org.primefaces.extensions.component.inputphone.InputPhone;
+
+/**
+ * Validator used with {@link InputPhone}.
+ *
+ * @author Jasper de Vries &lt;jepsar@gmail.com&gt;
+ */
+@FacesValidator("org.primefaces.extensions.validate.PhoneValidator")
```
jepsar: I could not find any validators in PF or PFE, so there isn't convention on how an ID should look like.

src/main/java/org/primefaces/extensions/validate/PhoneValidator.java
```diff
@@ -0,0 +1,59 @@
...
+            country = context.getExternalContext().getRequestParameterMap().get(inputPhone.getClientId() + "_iso2");
+            if (country == null || "auto".equals(country)) {
+                country = "";
+            }
+        }
+        try {
+            PhoneNumberUtil phoneNumberUtil = PhoneNumberUtil.getInstance();
+            Phonenumber.PhoneNumber phoneNumber = phoneNumberUtil.parse((String) object, country);
+            if (!phoneNumberUtil.isValidNumber(phoneNumber)) {
+                throw new ValidatorException(new FacesMessage("Invalid phone number"));
```
jepsar: I could not find strings that are loaded from a resource bundle in PF or PFE, so I hardcoded it for now. Is that okay? If so, I'll create a final static String as it is used twice.

src/main/java/org/primefaces/extensions/validate/PhoneValidator.java
```diff
@@ -0,0 +1,59 @@
...
+@FacesValidator("org.primefaces.extensions.validate.PhoneValidator")
+public class PhoneValidator implements Validator {
+
+    @Override
+    public void validate(FacesContext context, UIComponent component, Object object) throws ValidatorException {
+        String country = "";
+        if (InputPhone.class.isInstance(component)) {
+            InputPhone inputPhone = (InputPhone) component;
+            country = context.getExternalContext().getRequestParameterMap().get(inputPhone.getClientId() + "_iso2");
+            if (country == null || "auto".equals(country)) {
```
jepsar: To see if "auto" is set needs to be moved to PhoneInput. The renderer needs it as well. I'll change this.

src/main/java/org/primefaces/extensions/validate/PhoneValidator.java
```diff
@@ -0,0 +1,59 @@
...
+import org.primefaces.extensions.component.inputphone.InputPhone;
+
+/**
+ * Validator used with {@link InputPhone}.
+ *
+ * @author Jasper de Vries &lt;jepsar@gmail.com&gt;
+ */
+@FacesValidator("org.primefaces.extensions.validate.PhoneValidator")
+public class PhoneValidator implements Validator {
+
```
melloware, line 35: I don't see where you are actually using this validator?

jepsar (commented):
@melloware As agreed per chat I did a review with some questions.

src/main/resources/META-INF/resources/primefaces-extensions/inputphone/inputphone-widget.js
```diff
@@ -15,13 +15,14 @@ PrimeFaces.widget.ExtInputPhone = PrimeFaces.widget.BaseWidget.extend({
         this._super(cfg);
         this.id = cfg.id;
         this.cfg = cfg;
-        this.input = document.querySelector(this.jqId + "_input");
-        this.iti = intlTelInput(this.input, cfg);
 
-        this.inputJq = $(this.jqId + "_input");
+        this.inputJq = $(this.jqId + '_input');
         this.inputJq.data(PrimeFaces.CLIENT_ID_DATA, this.id);
         PrimeFaces.skinInput(this.inputJq);
```
melloware, line 23: why not bind `this.hiddenInput = $($this.jqId + '_iso2')` here as well.

src/main/resources/META-INF/resources/primefaces-extensions/inputphone/inputphone-widget.js
```diff
@@ -31,6 +32,7 @@ PrimeFaces.widget.ExtInputPhone = PrimeFaces.widget.BaseWidget.extend({
         this.input.addEventListener('countrychange', function () {
             if ($this.hasBehavior('countrySelect')) {
                 var country = $this.iti.getSelectedCountryData();
+                $($this.jqId + '_iso2').val(country.iso2);
```
melloware: Change to `$this.hiddenInput.val(country.iso2);`

melloware (changes requested):

pom.xml
melloware: If its Apache 2 we can add it.

src/main/java/org/primefaces/extensions/validate/PhoneValidator.java
melloware: I would like @tandraschko or @cnsgithub to weigh in but typically we don't make this optional we enforce validation in the rendering on Submit.

src/main/java/org/primefaces/extensions/validate/PhoneValidator.java
melloware: Good call. make it a constant.

src/main/java/org/primefaces/extensions/validate/PhoneValidator.java
cnsgithub: Basically, opt-in input validation means opt-in security as well, which isn't a good idea, is it?! :wink:

cnsgithub: However, just keep in mind, that implementations on client and server side may differ and result in unexpected behaviour, i.e. client side validation might pass but server side validation might not and vice versa.

src/main/java/org/primefaces/extensions/validate/PhoneValidator.java
jepsar: You would use it in the xhtml like `validator="org.primefaces.extensions.validate.PhoneValidator"`. See https://github.com/primefaces-extensions/core/pull/68#discussion_r265090394

src/main/java/org/primefaces/extensions/validate/PhoneValidator.java
cnsgithub: At least in PF there is a resource bundle _Messages_, see https://github.com/primefaces/primefaces/blob/96d30807fc67a2b6183f4956cb29202ae44759ee/src/main/java/org/primefaces/component/calendar/Calendar.java#L153 as an usage example.

src/main/java/org/primefaces/extensions/validate/PhoneValidator.java
jepsar: > Basically, opt-in input validation means opt-in security as well, which isn't a good idea, is it?!

It might depend on what you consider valid. For example: "1-800-pizza", is that a valid phone number? This validator should be a reasonable default I would say.

src/main/java/org/primefaces/extensions/validate/PhoneValidator.java
jepsar: Thanks, I'll have a look at that.

src/main/java/org/primefaces/extensions/validate/PhoneValidator.java
cnsgithub: There might of course be use-case-dependent validation rules. Nevertheless, if a phone number gets rejected on client side, it HAS TO BE rejected on server side as well. Therefore, it's essential for the client and server side implementations to be symmetric.

src/main/java/org/primefaces/extensions/validate/PhoneValidator.java
jepsar: @melloware Would that mean that we need to introduce something like `PFEMessages`?

src/main/java/org/primefaces/extensions/validate/PhoneValidator.java
jepsar: I don't think any checks are done on the client side. I'll see if we can have the server side to be as lenient as possible.

src/main/java/org/primefaces/extensions/validate/PhoneValidator.java
melloware: Yep fee free to introduce a PFE messages like PF does it.  I will then look through all the PFE code and convert things to use that messages.

melloware: Or if you want leave it hard coded for now, we can make it a clean up task for 7.1?

jepsar: @melloware Sounds like a bigger issue than this component alone. Let's create a new issue for this.

Update: https://github.com/primefaces-extensions/primefaces-extensions.github.com/issues/665

src/main/java/org/primefaces/extensions/validate/PhoneValidator.java
jepsar: Okay, I have added the validator by default.

jepsar:
@melloware @cnsgithub I've resolved the comments / requests above and did some minor improvements. Can you please have a look?

melloware (approved):
I like it!

pom.xml
```diff
@@ -437,6 +437,12 @@
 			<artifactId>poi-ooxml</artifactId>
 			<scope>provided</scope>
 		</dependency>
+        <!-- InputPhone dependencies -->
```
cnsgithub, line 440: Indentation

cnsgithub (changes requested):

PR #71, opened 2019-03-20
Fix #666 Added TimeAgo component

### LINKED ISSUE #666: CodeScanner: update zxing-js 0.18.6

jepsar:


src/main/java/org/primefaces/extensions/component/timeago/TimeAgo.java
```diff
@@ -0,0 +1,185 @@
...
+/**
+ * <code>TimeAgo</code> component.
+ *
+ * @author Jasper de Vries &lt;jepsar@gmail.com&gt;
+ * @since 7.0.1
+ */
+@ResourceDependencies({
+    @ResourceDependency(library = "primefaces", name = "jquery/jquery.js"),
+    @ResourceDependency(library = "primefaces", name = "core.js"),
+    @ResourceDependency(library = "primefaces-extensions", name = "timeago/jquery.timeago.js"),
```
jepsar, line 39: @melloware Can you compress these two JavaScripts? I'm still don't fully get how that plugin works.

src/main/java/org/primefaces/extensions/component/timeago/TimeAgo.java
```diff
@@ -0,0 +1,185 @@
...
+
+    public TimeAgo() {
+        setRendererType(DEFAULT_RENDERER);
+
+        final String i18nAbbr = getBundledLocaleAbbr(MessageFactory.getLocale());
+        if (i18nAbbr != null) {
+            ResourceUtils.addComponentResource(FacesContext.getCurrentInstance(),
+                                               "timeago/i18n/jquery.timeago." + i18nAbbr + ".js",
+                                               "primefaces-extensions",
+                                               "body");
```
jepsar, line 128: Is this okay? I would have preferred this to be last in the head, but I could not get it there (it ended up being loaded too soon).

jepsar (commented):
@melloware I think it's good enough for a first version. Can you have a look, please?

src/main/java/org/primefaces/extensions/component/timeago/TimeAgo.java
melloware: Yep I will take care of it

src/main/java/org/primefaces/extensions/component/timeago/TimeAgo.java
melloware: Instead of adding the JS here in the java code just pass the calculated locale to the Widget as a property.  Then in the Widget init() method you can check if locale != null then load the script in the widget like...

```javascript
if (locale) {
   var languageJsURI = PrimeFaces.resources.getFacesResource('/timeago/i18n/jquery.timeago.' + locale+'.js',
                    PrimeFacesExt.RESOURCE_LIBRARY, PrimeFacesExt.VERSION);
    // load script
    PrimeFacesExt.getScript(languageJsURI , $.proxy(function(data, textStatus) {
                    this.timeAgo = $(this.selector).timeago();
                }, this));
} else {
     this.timeAgo = $(this.selector).timeago();
}

```

melloware (changes requested):

melloware:
I merged it I will make some cleanup changes then you can fix the JS loading after that.

melloware:
OK I checked in some changes I have the JavaScript combining properly now to timeago.js and updated your Component to respect that.  The IL8N files stay there untouched in that directory.  Now you should be good to work on moving that IL8N loading to the widget init!

PR #82, opened 2019-06-15
Fix #705: add extender attribute to GChart component.

fheldt:
This implements the changes needed for issue #705.

### DIFF
--- a/src/main/java/org/primefaces/extensions/component/gchart/GChartRenderer.java
+++ b/src/main/java/org/primefaces/extensions/component/gchart/GChartRenderer.java
@@ -74,7 +74,9 @@ public class GChartRenderer extends CoreRenderer {
                     .attr("language", chart.getLanguage())
                     .attr("width", chart.getWidth())
                     .attr("height", chart.getHeight());
-
+        if (chart.getExtender() != null) {
+            wb.nativeAttr("extender", chart.getExtender());
+        }
         encodeClientBehaviors(context, chart);
 
         wb.finish();
--- a/src/main/resources/META-INF/resources/primefaces-extensions/gchart/gchart.js
+++ b/src/main/resources/META-INF/resources/primefaces-extensions/gchart/gchart.js
@@ -39,6 +39,15 @@ PrimeFaces.widget.ExtGChart = PrimeFaces.widget.BaseWidget.extend({
             });
         });
 
+        // user extension to configure gchart
+        var extender = this.cfg.extender;
+        if (extender) {
+            if (typeof extender === "function") {
+                extender.call(this);
+            } else {
+                PrimeFaces.error("Extender value is not a javascript function!");
+            }
+        }
     },
 
     draw : function() {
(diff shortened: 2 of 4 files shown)


src/main/java/org/primefaces/extensions/component/gchart/GChartRenderer.java
```diff
@@ -74,7 +74,9 @@ protected void encodeScript(final FacesContext context, final GChart chart) thro
                     .attr("language", chart.getLanguage())
                     .attr("width", chart.getWidth())
                     .attr("height", chart.getHeight());
-
+        if (chart.getExtender() != null) {
+            wb.nativeAttr("extender", chart.getExtender());
```
jepsar, line 78: If the extender value is just the name of the function, I think you should just use `wb.attr`

src/main/resources/META-INF/resources/primefaces-extensions/gchart/gchart.js
```diff
@@ -39,6 +39,15 @@ PrimeFaces.widget.ExtGChart = PrimeFaces.widget.BaseWidget.extend({
             });
         });
 
+        // user extension to configure gchart
+        var extender = this.cfg.extender;
+        if (extender) {
+            if (typeof extender === "function") {
```
jepsar, line 45: Oh, it's not the name of the function.. it _is_ the function. Can you please clarify that in the attribute description?

src/main/java/org/primefaces/extensions/component/gchart/GChartRenderer.java
melloware: I believe native attribute is correct because it’s a reference to a JavaScript function.

jepsar: Yeah, I was confused by the attribute documentation (see below).

src/main/resources/META-INF/resources/primefaces-extensions/gchart/gchart.js
fheldt: Sorry, I'm not quite sure how to clarify the attribute description. I used the same description used in primefaces (eg schedule component) or primefaces-extensions (sheet component). It is exactly that: the name of a Javascript function which is called during initialization.

fheldt: Just to give a small code snippet on how to use this:
```
	<script src="https://www.gstatic.com/charts/loader.js"></script>
	<h:outputScript>
		function extendChart() {
			// this = widget
			console.log("extendChart called...");
			this.options.gantt = {
				criticalPathEnabled: true,
				criticalPathStyle: {
					stroke: '#e64a19',
					strokeWidth: 5
				}
			}
		}
	</h:outputScript>
	<h:form>
		<pe:gChart id="gantt" value="#{ganttChartCtrl.chart}" title="GANTT" extender="extendChart">
			<p:ajax listener="#{ganttChartCtrl.onSelect}"/>
		</pe:gChart>
	</h:form>
```

PR #84, opened 2019-07-26
Fix #713 Sheet: Filtering and dynamic columns not working together

blutorange:
# Changes 

This should fix issue #713. I made the following changes:

* I added some clarification to the tag documentation for the `sortBy` attribute of the sheet. It specifies the property by which the sheet should be sorted initially. 
* `origSortBy` used to store the original value expression for resetting the sort. The part of the code where this is set was commented out, so it does not seem to have had any effect. I changed it `currentSortBy`, which now saves the ID of the column by which the sheet is currently sorted. The attribute `sortBy` is never modified, so there is no need to reset it.
* I also tested this commit with `movableCols` set to `true`. When the sheet is sorted by say the 3rd column, and the user moves the 2nd column to the right, the 3rd column is not at index 2. The sheet still shows the 3rd column as the sorted column. But this does not work in the current version of PF extensions either, so this is a different isues/ticket.
* When resetting the sheet, `currentSortBy` is simply reset to `null`. As mentioned above, storing the old value `origSortBy` was commented out. Was this intentional? I think this would remove any sorting when the sheet is reset. In case that was intentional, you should add the following line to `Sheet#resetSort` to remove any sorting on reset:

```java
getStateHelper().put(PropertyKeys.sortBy.name(), null);
```

To determine by which column to sort, I changed the code as follows:

1. Check if a valid column ID was stored in `currentSortBy`. If so, sort by that.
1. Otherwise, check if the `sortBy` expression evaluates to a valid ID of a column. If so, sort by that.
1. Otherwise, check if the literal value expression stored in the `sortBy` attribute matches the value expression of a sheet column. If so, use that.
1. Otherwise, if the `sortBy` attribute of the sheet is a value expression, sort the sheet data by that property (even when no column specifies a `sortBy` attribute)

The last two cases are there to ensure compatibility with xhtml code for previous versions. In case you do not require this compatibility, the code could be simplified a bit.

I think this is easier to understand by giving a few examples of how it works now.

This still works:

```xml
<pe:sheet id="sheet" value="#{sheetController.assets}" var="row" 
    sortBy="#{row.assetId}" sortOrder="ascending" >
    <pe:sheetcolumn sortBy="#{row.assetId}"  />
    <pe:sheetcolumn sortBy="#{row.assetType}"  />
</pe:sheet>
```

but you can now also write:

```xml
<pe:sheet id="sheet" value="#{sheetController.assets}" var="row" 
    sortBy="assertId" sortOrder="ascending" >
    <pe:sheetcolumn id="assetId" sortBy="#{row.assetId}"  />
    <pe:sheetcolumn id="assetType" sortBy="#{row.assetType}"  />
</pe:sheet>
```

The latter way is the only way to make it work for dynamic columns (sort by the 3rd column initially):

```xml
<pe:sheet id="sheet" value="#{sheetDynamicController.sheetRows}" var="row"
    sortBy="col2" sortOrder="ascending">
    <c:forEach items="#{sheetDynamicController.hoursOfDay}" var="hourOfDay" 
        varStatus="status">
        <pe:sheetcolumn id="col#{status.index}" sortBy="#{row.cells[status.index].value}" />
    /c:forEach>
</pe:sheet>
```

And in the above example, the user can also click on any column and it sorts by that column, which fixes issue #713:

image

Finally, you can also omit the `sortBy` attribute on sheet column and only specify `sortBy` on the sheet, which just performs some initial sorting of the data:

```xml
<pe:sheet id="sheet" value="#{sheetAjaxController.assets}"
    sortBy="#{row.assetId}" sortOrder="ascending">
	<pe:sheetcolumn headerText="Id (readOnly)" value="#{row.assetId}" />
	<pe:sheetcolumn headerText="Type (readOnly)" value="#{row.assetType}" />
	<pe:sheetcolumn headerText="Platform (readOnly)" value="#{row.platform}" />
	<pe:sheetcolumn headerText="Arch (readOnly)" value="#{row.platformArch}" />
	<pe:sheetcolumn headerText="Editable" value="#{row.value1}" colType="numeric"/>
</pe:sheet>
```

### DIFF
--- a/src/main/java/org/primefaces/extensions/component/sheet/Sheet.java
+++ b/src/main/java/org/primefaces/extensions/component/sheet/Sheet.java
@@ -230,7 +230,7 @@ public class Sheet extends SheetBase {
      */
     public void resetSort() {
         // Set to null to restore initial sort order specified by sortBy
-        getStateHelper().put(PropertyKeys.currentSortBy, null);
+        getStateHelper().put(PropertyKeys.currentSortBy.name(), null);
 
         final String origSortOrder = (String) getStateHelper().get(PropertyKeys.origSortOrder);
         if (origSortOrder != null) {
--- a/src/main/java/org/primefaces/extensions/component/sheet/Sheet.java
+++ b/src/main/java/org/primefaces/extensions/component/sheet/Sheet.java
@@ -229,9 +229,8 @@ public class Sheet extends SheetBase {
      * Resets the sorting to the originally specified values (if any)
      */
     public void resetSort() {
-        final ValueExpression origSortBy = (ValueExpression) getStateHelper().get(PropertyKeys.origSortBy);
-        // Set sort by even if null to restore to initial sort order.
-        setSortByValueExpression(origSortBy);
+        // Set to null to restore initial sort order specified by sortBy
+        getStateHelper().put(PropertyKeys.currentSortBy, null);
 
         final String origSortOrder = (String) getStateHelper().get(PropertyKeys.origSortOrder);
         if (origSortOrder != null) {
@@ -262,7 +261,8 @@ public class Sheet extends SheetBase {
     /**
      * Updates a submitted value.
      *
-     * @param row
+     * @param context
+     * @param rowKey
      * @param col
      * @param value
      */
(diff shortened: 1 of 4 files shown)


melloware (approved):
Great PR!

PR #86, opened 2019-08-24
Fix 701: Added keepOpen attribute to FAB

jepsar:
@melloware sorry for the delay, I didn't manage to find the time to finish both issues during the week.

Can you please have a look at this PR. I've tested it, the JS is working, but I can't get the attribute to be true. I must be overlooking something trivial, but I can't find what.

### DIFF
--- a/src/main/java/org/primefaces/extensions/component/fab/FloatingActionButtonRenderer.java
+++ b/src/main/java/org/primefaces/extensions/component/fab/FloatingActionButtonRenderer.java
@@ -109,6 +109,9 @@ public class FloatingActionButtonRenderer extends BaseMenuRenderer {
         final FloatingActionButton fab = (FloatingActionButton) menu;
         final WidgetBuilder wb = getWidgetBuilder(context);
         wb.init("ExtFAB", fab.resolveWidgetVar(), fab.getClientId(context));
+        if (fab.isKeepOpen()) {
+            wb.attr("keepOpen", fab.isKeepOpen());
+        }
         wb.finish();
     }
 
--- a/src/main/resources/META-INF/primefaces-extensions.taglib.xml
+++ b/src/main/resources/META-INF/primefaces-extensions.taglib.xml
@@ -6818,7 +6818,7 @@ See showcase for an example. Default null.]]>
             </description>
             <name>keepOpen</name>
             <required>false</required>
-            <type>boolean</type>
+            <type>java.lang.Boolean</type>
         </attribute>
         <attribute>
             <description>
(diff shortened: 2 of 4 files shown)


src/main/java/org/primefaces/extensions/component/fab/FloatingActionButtonRenderer.java
```diff
@@ -109,6 +109,9 @@ protected void encodeScript(FacesContext context, AbstractMenu menu) throws IOEx
         final FloatingActionButton fab = (FloatingActionButton) menu;
         final WidgetBuilder wb = getWidgetBuilder(context);
         wb.init("ExtFAB", fab.resolveWidgetVar(), fab.getClientId(context));
+        if (fab.isKeepOpen()) {
```
jepsar, line 112: @melloware somehow this always returns false.. I don't understand why

src/main/resources/META-INF/primefaces-extensions.taglib.xml
```diff
@@ -6812,6 +6812,14 @@ See showcase for an example. Default null.]]>
             <required>false</required>
             <type>java.lang.String</type>
         </attribute>
+        <attribute>
+            <description>
+                <![CDATA[Keep the menu open when an item is clicked. Default value is false.]]>
+            </description>
+            <name>keepOpen</name>
+            <required>false</required>
+            <type>boolean</type>
```
melloware: Change "boolean" to "java.lang.Boolean" and it will work!

src/main/resources/META-INF/resources/primefaces-extensions/fab/fab-widget.js
```diff
@@ -25,8 +25,13 @@ PrimeFaces.widget.ExtFAB = PrimeFaces.widget.BaseWidget.extend({
                 document.selection.empty();
             }
         };
-        var that = this;
-        this.button.click(function(){ that.toggle(); });
+        var obj = this;
+        this.button.click(function(){ obj.toggle(); });
+        if (cfg.keepOpen === undefined) {
```
melloware: I think you can say "if (!cg.keepOpen)" which will evaluate if its FALSE or UNDEFINED?

melloware (changes requested):

src/main/java/org/primefaces/extensions/component/fab/FloatingActionButtonRenderer.java
jepsar: Oh no! My test was wrong.. I used the attribute on the menu item!

src/main/resources/META-INF/resources/primefaces-extensions/fab/fab-widget.js
jepsar: Correct!

src/main/resources/META-INF/primefaces-extensions.taglib.xml
jepsar: Both are used in PFE. I don't think it really matters in this case. Changed it though.

melloware (approved):

PR #89, opened 2019-10-02
Fixes #732 by attaching event listeners after a mode change

### LINKED ISSUE #732: CKEditor: Events not fired anymore after changing the mode
To see this, change the `Basic Usage` showcase for the CKEditor like this by adding an AJAX listener for the `change` event:

image

```xml
<p:growl id="growl" showDetail="true" />  
  
<pe:ckEditor id="editor" value="#{editorController.content}" interfaceColor="#{editorController.color}">  
    <p:ajax event="save" listener="#{editorController.saveListener}" update="growl"/>  
    <p:ajax event="change" onstart="console.log('change')"/>  
</pe:ckEditor>  
  
<p:commandButton actionListener="#{editorController.changeColor}" update="@form:centerExample"  
                 value="Change color with AJAX" style="margin-top:10px;"/>  
```

When you type anything in the CKEditor, the message `change` is logged. But if you, for example, switch to the source code mode and then switch back, it won't work anymore.

---

The reason for this lies in the way the event listeners are attached to the CKEditor in the widget. It's explained [here on the CKEditor docs](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_editable.html#method-attachListener) and also [here on stackexchange](https://stackoverflow.com/a/22443086).

Events are currently attached like this:

```javascript
        this.instance.editable().attachListener(editable, 'paste', $.proxy(function(event) {
                this.checkChange();
                this.checkDirty();
        }, this));
```

 The `editable` refers to the DOM element of the editor content. This element is replaced, for exapmle, when you change the mode of the CKEditor. As mentioned in the docs, this should probably be changed to

```javascript
        //changes to WYSIWYG mode
    	this.instance.on('contentDom', $.proxy(function(event) {
            this.bindEditorEvents(this.instance);
            this.fireEvent('wysiwygMode');
      	}, this));
```

with the `bindEditorEvents` method containing the parts where the events listeners are attached to the `editor.editable()` instance:

Also, `contentDom` is fired only when the editor is switch to WYSIWYG mode. To make the events work in source code mode, [they can be bound like this](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_editor.html#event-change):

```javascript
    	//changes to source mode
    	this.instance.on('mode', $.proxy(function(event) {
                if (this.instance.mode != 'source') {
                    return;
                }
                this.bindEditorEvents(this.instance);
                this.fireEvent('sourceMode');
        }, this));
```

---

I already tried this locally, if you want I can make a PR for this.

blutorange:
(If you use the github UI to view the diff changes, apparently you can add the URL parameter `w=1` to ignore whitespaces, which helps a lot here)

This fixes primefaces-extensions/primefaces-extensions.github.com#732 and primefaces-extensions/primefaces-extensions.github.com#734 . I'll take a look at adding an example to the showcase for the events. 

* The AJAX behavior events `dirty` and 'change' are fired now even after a mode change, and also when the CKEdItor is in the source mode.
* Listen to the `contentDom` event that is triggered when the WYSIWYG editable DOM element was replaced, and reattach the required event listeners.
* Listen to the `mode` event that is triggered whenever the mode changes. When the mode is changed to `source`, attach the event listeners as well to trigger the behavior events in source mode too.
* Listen to the `keyup` event instead of the `keydown` event - this allows the `dirty` AJAX behavior event to fire on the first keypress.
* Indentation was a mix between spaces and tabs. I replaced all tabs with whitespaces (for the CKEditor widget), as this is how it's done in most other widgets.

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces-extensions/ckeditor/widget.js
+++ b/src/main/resources/META-INF/resources/primefaces-extensions/ckeditor/widget.js
@@ -16,7 +16,7 @@ CKEDITOR_GETURL = function(resource) {
             };
         }
     }
-	
+    
     //do not resolve
     if (resource.indexOf('?resolve=false') !== -1) {
         facesResource = resource.replace('?resolve=false', '');
@@ -89,12 +89,12 @@ CKEDITOR_GETURL = function(resource) {
  */
 PrimeFaces.widget.ExtCKEditor = PrimeFaces.widget.DeferredWidget.extend({
 
-	/**
-	 * Initializes the widget.
-	 *
-	 * @param {object} cfg The widget configuration.
-	 */
-	init : function(cfg) {
+    /**
+     * Initializes the widget.
+     *
+     * @param {object} cfg The widget configuration.
+     */
+    init : function(cfg) {
         this._super(cfg);
 
         this.instance = null;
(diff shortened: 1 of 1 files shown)


melloware:
Looks good thanks for the PR!

PR #90, opened 2019-10-03
Fix #735, keep scroll position of CKEditor and CKMirror

blutorange:
This solves primefaces-extensions/primefaces-extensions.github.com#735. CodeMirror was easy, but CKEditor is pretty complex. I added checks for the various properties of CKEditor to make sure we don't get any JavaScript errors when some of its elements are `undefined`.

* The `refresh` hook of the widgets is (sometimes) only called after the previous element was removed. So instead, we register a listener on the `scroll` event and save the current scroll position.
* Then, when the CKEditor or CodeMirror is updated via AJAX, we can restore the scroll position.
* CKEditor has got two modes, `source` and `WYSIWYG`. Depending on the mode, the scroll listener needs to be registered differently. When the editor was in source mode before an AJAX update, we also set it back to source mode after the update. Also, care needs to be taken to avoid registering a listener twice.


melloware:
Very nice job.

PR #91, opened 2019-10-09
Fix 723 Cleaned up TimePicker styling

### LINKED ISSUE #723: TimePicker: doesn't look okay with modern themes
Go to https://www.primefaces.org/showcase-ext/sections/timePicker/basicUsage.jsf and switch to nova-light for example.

jepsar:
This should fix https://github.com/primefaces-extensions/primefaces-extensions.github.com/issues/723

It's looking good on my MacBook with Chrome. We might want to test with some other browsers. I think we are alright though, since the majority of the styling is inherited from existing components (like spinner and input group)

### DIFF
--- a/src/main/java/org/primefaces/extensions/component/timepicker/TimePicker.java
+++ b/src/main/java/org/primefaces/extensions/component/timepicker/TimePicker.java
@@ -23,7 +23,6 @@ import java.util.HashMap;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Map.Entry;
-
 import javax.faces.application.ResourceDependencies;
 import javax.faces.application.ResourceDependency;
 import javax.faces.component.html.HtmlInputText;
@@ -31,7 +30,6 @@ import javax.faces.context.FacesContext;
 import javax.faces.event.AjaxBehaviorEvent;
 import javax.faces.event.FacesEvent;
 import javax.faces.event.PhaseId;
-
 import org.primefaces.component.api.Widget;
 import org.primefaces.extensions.event.BeforeShowEvent;
 import org.primefaces.extensions.event.CloseEvent;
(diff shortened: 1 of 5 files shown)


melloware:
I will check this out on Edge, IE and Firefox and let you know how it looks!

melloware:
This looks good.  There are still a few issues though...

1. Look at the Spinners in Nova-Light and Nova-Dark.
image

2. The popup styling is off in the premium themes.  
**Good (Omega):**
image

**Bad (Nova-Dark):**
image

melloware (changes requested):
See comment in conversation.

jepsar:
The customization input is an issue with the showcase (the with of that input is hardcoded there). We might want to add 10 to 15px there.

What browser gave you issues with the spinners?

What is wrong with the popup (other than the button style)?

melloware:
I was testing in Chrome and Firefox.  Didn't try IE and Edge yet.  I am on Windows.

As for the popup doesn't it look really naked compared to Omega? No borders around dates etc?  Maybe its supposed to be that way it just looks awfully plain and the buttons are not styled.

image

vs Omega...

image

melloware:
OK the good news is its working in Firefox, Chrome, IE and Edge.  But the the same spinner and popup rendering is consistent in all 4 browsers.

jepsar:
> As for the popup doesn't it look really naked compared to Omega?

I tried to apply very mild styling and to keep it in line with, for example, `p:calendar`

> But the the same spinner and popup rendering is consistent in all 4 browsers.

Okay, I'll look in to that.

jepsar:
@melloware can you please test again? Please note that I've also submitted a change to the showcase: https://github.com/primefaces-extensions/showcase/pull/22

melloware:
I will check it out!

PR #92, opened 2019-11-06
Support LocalDateTime and ZonedDateTime (and some cleanup)

jepsar:
This is the first half related to https://github.com/primefaces-extensions/primefaces-extensions.github.com/issues/677

### DIFF
--- a/src/main/java/org/primefaces/extensions/component/timeago/TimeAgo.java
+++ b/src/main/java/org/primefaces/extensions/component/timeago/TimeAgo.java
@@ -16,17 +16,17 @@
 package org.primefaces.extensions.component.timeago;
 
 import java.text.SimpleDateFormat;
-import java.util.Arrays;
+import java.time.LocalDateTime;
+import java.time.ZoneId;
+import java.time.ZonedDateTime;
+import java.time.format.DateTimeFormatter;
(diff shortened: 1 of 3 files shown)


src/main/java/org/primefaces/extensions/component/timeago/TimeAgo.java
```diff
@@ -52,67 +52,6 @@
 
     private static final String DATE_FORMAT = "yyyy-MM-dd'T'HH:mm:ssZ";
 
-    private static final List<String> BUNDLED_LOCALES = Arrays.asList(
```
jepsar, line 55: This should have been removed in the previous fix where we get the translations from Javascript, so I included it in this PR.

src/main/java/org/primefaces/extensions/component/timeago/TimeAgo.java
```diff
@@ -190,35 +129,41 @@ public Locale calculateLocale() {
         return appropriateLocale;
     }
 
-    public final String getBundledLocale() {
```
jepsar, line 193: See previous remark.

melloware (approved):

melloware:
Looks great!

PR #93, opened 2019-11-28
Fixed countries string to list

jepsar:
@melloware I think this was a bug. It should have been converting `"nl be de"` to `["nl", "be", "de"]` (it should support both comma and space separated). I did this using the online editor in Git, but I think I need to give this a second look.


melloware (approved):

melloware:
I like the fix.  Merging.

PR #95, opened 2020-02-23
Added CommandButtonSingleClickRenderer

### LINKED ISSUE #760: Add single click renderer for CommandButton
See https://github.com/primefaces/primefaces/issues/2521

jepsar:
Fix https://github.com/primefaces-extensions/primefaces-extensions.github.com/issues/760

See also https://github.com/primefaces-extensions/showcase/pull/23

### DIFF
--- a/src/main/java/org/primefaces/extensions/renderer/CommandButtonSingleClickRenderer.java
+++ b/src/main/java/org/primefaces/extensions/renderer/CommandButtonSingleClickRenderer.java
@@ -16,7 +16,6 @@
 package org.primefaces.extensions.renderer;
 
 import java.io.IOException;
-import java.util.stream.Stream;
 import javax.faces.component.UIComponent;
 import javax.faces.context.FacesContext;
 import org.primefaces.component.commandbutton.CommandButton;
@@ -30,8 +29,6 @@ import org.primefaces.component.commandbutton.CommandButtonRenderer;
  */
 public class CommandButtonSingleClickRenderer extends CommandButtonRenderer {
 
-    protected static final String[] CLASS_CONFIRM = {"ui-confirmdialog-yes", "ui-confirmdialog-no"};
-
     @Override
     public void encodeEnd(FacesContext context, UIComponent component) throws IOException {
         CommandButton button = (CommandButton) component;
@@ -44,8 +41,10 @@ public class CommandButtonSingleClickRenderer extends CommandButtonRenderer {
     }
 
     protected boolean isConfirmation(final CommandButton button) {
-        return button.getStyleClass() != null
-               && Stream.of(CLASS_CONFIRM).anyMatch(s -> s.contains(button.getStyleClass()));
+        String styleClass = button.getStyleClass();
+        return styleClass != null
+               && (styleClass.contains("ui-confirmdialog-yes")
+                   || styleClass.contains("ui-confirmdialog-no"));
     }
 
     protected String toggle(final String widgetVar, final boolean enabled) {
--- a/src/main/java/org/primefaces/extensions/renderer/CommandButtonSingleClickRenderer.java
+++ b/src/main/java/org/primefaces/extensions/renderer/CommandButtonSingleClickRenderer.java
@@ -21,7 +21,7 @@ import org.primefaces.component.commandbutton.CommandButton;
 import org.primefaces.component.commandbutton.CommandButtonRenderer;
 
 /**
- * {@link CommandButton} renderer disabling the button while action is processed.
+ * {@link CommandButton} renderer disabling the button while action is processed for buttons that are using Ajax.
  *
  * @author Jasper de Vries &lt;jepsar@gmail.com&gt;
  * @since 8.0
@@ -30,12 +30,19 @@ public class CommandButtonSingleClickRenderer extends CommandButtonRenderer {
 
     @Override
     protected void encodeMarkup(FacesContext context, CommandButton button) throws IOException {
-        if (button.isRendered() && !button.isDisabled() && !isConfirmation(button)) {
+        if (isToggleable(button)) {
             String widgetVar = button.resolveWidgetVar(context);
             button.setOnclick(prefix(button.getOnclick(), toggle(widgetVar, false)));
             button.setOncomplete(prefix(button.getOncomplete(), toggle(widgetVar, true)));
         }
         super.encodeMarkup(context, button);
+    }
+
+    protected boolean isToggleable(final CommandButton button) {
+        if (!button.isRendered() || button.isDisabled() || isConfirmation(button)) {
+            return false;
+        }
+        return button.isAjax();
     }
 
     protected boolean isConfirmation(final CommandButton button) {
--- a/src/main/java/org/primefaces/extensions/renderer/CommandButtonSingleClickRenderer.java
+++ b/src/main/java/org/primefaces/extensions/renderer/CommandButtonSingleClickRenderer.java
@@ -16,7 +16,6 @@
 package org.primefaces.extensions.renderer;
 
 import java.io.IOException;
-import javax.faces.component.UIComponent;
 import javax.faces.context.FacesContext;
 import org.primefaces.component.commandbutton.CommandButton;
 import org.primefaces.component.commandbutton.CommandButtonRenderer;
@@ -30,14 +29,13 @@ import org.primefaces.component.commandbutton.CommandButtonRenderer;
 public class CommandButtonSingleClickRenderer extends CommandButtonRenderer {
 
     @Override
-    public void encodeEnd(FacesContext context, UIComponent component) throws IOException {
-        CommandButton button = (CommandButton) component;
+    protected void encodeMarkup(FacesContext context, CommandButton button) throws IOException {
         if (button.isRendered() && !button.isDisabled() && !isConfirmation(button)) {
             String widgetVar = button.resolveWidgetVar(context);
             button.setOnclick(prefix(button.getOnclick(), toggle(widgetVar, false)));
             button.setOncomplete(prefix(button.getOncomplete(), toggle(widgetVar, true)));
         }
-        super.encodeEnd(context, component);
+        super.encodeMarkup(context, button);
     }
 
     protected boolean isConfirmation(final CommandButton button) {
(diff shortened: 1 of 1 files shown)


src/main/java/org/primefaces/extensions/renderer/CommandButtonSingleClickRenderer.java
```diff
@@ -0,0 +1,59 @@
...
+        super.encodeEnd(context, component);
+    }
+
+    protected boolean isConfirmation(final CommandButton button) {
+        return button.getStyleClass() != null
+               && Stream.of(CLASS_CONFIRM).anyMatch(s -> s.contains(button.getStyleClass()));
+    }
+
+    protected String toggle(final String widgetVar, final boolean enabled) {
+        return String.format("PF('%s').%sable();", widgetVar, enabled ? "en" : "dis");
```
Rapster: That line put a smile on my face, this is original! Though, it's `able()` and not `sable()` 😉

src/main/java/org/primefaces/extensions/renderer/CommandButtonSingleClickRenderer.java
```diff
@@ -0,0 +1,59 @@
...
+            String widgetVar = button.resolveWidgetVar(context);
+            button.setOnclick(prefix(button.getOnclick(), toggle(widgetVar, false)));
+            button.setOncomplete(prefix(button.getOncomplete(), toggle(widgetVar, true)));
+        }
+        super.encodeEnd(context, component);
+    }
+
+    protected boolean isConfirmation(final CommandButton button) {
+        return button.getStyleClass() != null
+               && Stream.of(CLASS_CONFIRM).anyMatch(s -> s.contains(button.getStyleClass()));
```
Rapster: I don't think the condition is right, I think you mean `button.getStyleClass().contains(s)`. Being said, instead of creating a stream everytime, I would have write 2 conditions instead:

```java
protected boolean isConfirmation(final CommandButton button) {
        String styleClass = button.getStyleClass();
        return styleClass != null
               && (styleClass.contains("ui-confirmdialog-yes") 
               || styleClass.contains("ui-confirmdialog-no"));
}
```

Why is this necessary btw?

src/main/java/org/primefaces/extensions/renderer/CommandButtonSingleClickRenderer.java
```diff
@@ -0,0 +1,59 @@
...
+ *
+ * @author Jasper de Vries &lt;jepsar@gmail.com&gt;
+ * @since 8.0
+ */
+public class CommandButtonSingleClickRenderer extends CommandButtonRenderer {
+
+    protected static final String[] CLASS_CONFIRM = {"ui-confirmdialog-yes", "ui-confirmdialog-no"};
+
+    @Override
+    public void encodeEnd(FacesContext context, UIComponent component) throws IOException {
```
Rapster: Better to override `encodeMarkup` instead no?

src/main/java/org/primefaces/extensions/renderer/CommandButtonSingleClickRenderer.java
jepsar: 😄 the `%s` is the placeholder, so it's correct.

src/main/java/org/primefaces/extensions/renderer/CommandButtonSingleClickRenderer.java
jepsar: Else the confirm dialog buttons will be disabled on click. You are right about the condition being inverted. I'll change it to what you suggested.

src/main/java/org/primefaces/extensions/renderer/CommandButtonSingleClickRenderer.java
jepsar: Why? I don't think it'll make much of a difference.

Rapster: It doesn't indeed, but `encodeMarkup` is meant for HTML stuff

src/main/java/org/primefaces/extensions/renderer/CommandButtonSingleClickRenderer.java
Rapster: Oh yeah indeed! My bad!

src/main/java/org/primefaces/extensions/renderer/CommandButtonSingleClickRenderer.java
jepsar: We are modifying attributes and don't really render anything.

Rapster: exactly, HTML attributes

jepsar: Alright, done. Also saved one line 😄

melloware (approved):
This is nice and clean!

Rapster (approved):

jepsar:
@melloware can you have a look at the showcase as well please?

https://github.com/primefaces-extensions/showcase/pull/23

Rapster:
But we'll still have the same problem with non ajax command no? Let's say I have a button which validate a form and does a redirect, if form not correctly filled I'll have a validation error and my button will remain disable (since onComplete not called)

melloware:
Maybe we need the enable logic onError too? to renable the button?

jepsar:
Alright, let me have a look at that.

jepsar:
@Rapster I'd suggest to add an Ajax check and only inject the JS for buttons using Ajax. I don't see how we can solve enabling the button for non-Ajax requests, other than having the user to implement something to enable the button again or adding a time out.

@melloware `oncomplete` is also called on error I believe.

melloware:
Agreed for now we should only add this to Ajax buttons.

src/main/java/org/primefaces/extensions/renderer/CommandButtonSingleClickRenderer.java
```diff
@@ -0,0 +1,63 @@
...
+            return false;
+        }
+        return button.isAjax();
+    }
+
+    protected boolean isConfirmation(final CommandButton button) {
+        String styleClass = button.getStyleClass();
+        return styleClass != null
+               && (styleClass.contains("ui-confirmdialog-yes")
+                   || styleClass.contains("ui-confirmdialog-no"));
```
melloware: What about just one check for styleClass.contains("ui-confirmdialog") ?

melloware (changes requested):

src/main/java/org/primefaces/extensions/renderer/CommandButtonSingleClickRenderer.java
jepsar: I considered that.. bit cleaner that way, right?

melloware (approved):

src/main/java/org/primefaces/extensions/renderer/CommandButtonSingleClickRenderer.java
```diff
@@ -0,0 +1,61 @@
...
+    protected void encodeMarkup(FacesContext context, CommandButton button) throws IOException {
+        if (isToggleable(button)) {
+            String widgetVar = button.resolveWidgetVar(context);
+            button.setOnclick(prefix(button.getOnclick(), toggle(widgetVar, false)));
+            button.setOncomplete(prefix(button.getOncomplete(), toggle(widgetVar, true)));
+        }
+        super.encodeMarkup(context, button);
+    }
+
+    protected boolean isToggleable(final CommandButton button) {
```
Rapster: I don't understand what you mean by `isToggleable`? Also, you can simplify your code with:
```java
return button.isRendered() && !button.isDisabled() && !isConfirmation(button) && button.isAjax()
```

jepsar: `isToggleable` will tell you if we can toggle enabled / disabled on it. I moved it to a method to prevent an if with the mother of all conditions. I think this is more readable.

What name would be better? needsListeners? disabledStateShouldBeToggled?

melloware: @rapster if you have a better name we are all ears!

Rapster: I don't know @melloware, I only asked to understand but maybe something like `isEligible`/`isEligibleToSingleClick`

src/main/java/org/primefaces/extensions/renderer/CommandButtonSingleClickRenderer.java
```diff
@@ -0,0 +1,61 @@
...
+        return button.isAjax();
+    }
+
+    protected boolean isConfirmation(final CommandButton button) {
+        String styleClass = button.getStyleClass();
+        return styleClass != null && styleClass.contains("ui-confirmdialog");
+    }
+
+    protected String toggle(final String widgetVar, final boolean enabled) {
+        return String.format("PF('%s').%sable();", widgetVar, enabled ? "en" : "dis");
```
melloware: So what abou the scenario in which the Button that was clicked updates the Button so it is no longer rendered.  When the onComplete fires won't the widget not exist to call Enable on?  So I think your original script may be more correct like..
```javascript
var w = PF('%s');if (w){w.%sable();}
```

melloware (changes requested):

src/main/java/org/primefaces/extensions/renderer/CommandButtonSingleClickRenderer.java
jepsar: Yeah.. I was thinking of why I had those null checks there.. just to be safe, or was I running into issues. Now you are pointing this out I think the latter. I'll change this tonight (CET).

src/main/java/org/primefaces/extensions/renderer/CommandButtonSingleClickRenderer.java
jepsar: `isEligible` it is!

src/main/java/org/primefaces/extensions/renderer/CommandButtonSingleClickRenderer.java
jepsar: Indeed, this check is required (see also update of the showcase)

melloware (approved):
This was great teamwork!

jepsar:
I think we are good to merge, right? @Rapster ?

PR #96, opened 2020-03-10
Fix 761 - Store and use the original attribute values

### LINKED ISSUE #761: CommandButtonSingleClick: Problem in Datatable
If you have a Datatable with a CommandButton in the first column.

```xml
<p:column>
    <p:commandButton id="btnActions" value="Button" 
         oncomplete="alert('#{component.clientId}');" immediate="true" process="@this" />
<p:column>
```

Because in encodeMarkup we are calling button.resolvewidgetVar() it is resolving it incorrectly.

All of the buttons in my table are being rendered with widgetVar :0:

```
frmMain:tblTest:0:btnActions
frmMain:tblTest:0:btnActions
frmMain:tblTest:0:btnActions
frmMain:tblTest:0:btnActions
```

instead of 

```
frmMain:tblTest:0:btnActions
frmMain:tblTest:1:btnActions
frmMain:tblTest:2:btnActions
frmMain:tblTest:3:btnActions
```

jepsar:
Fix https://github.com/primefaces-extensions/primefaces-extensions.github.com/issues/761

### DIFF
--- a/src/main/java/org/primefaces/extensions/renderer/CommandButtonSingleClickRenderer.java
+++ b/src/main/java/org/primefaces/extensions/renderer/CommandButtonSingleClickRenderer.java
@@ -19,6 +19,7 @@ import java.io.IOException;
 import javax.faces.context.FacesContext;
 import org.primefaces.component.commandbutton.CommandButton;
 import org.primefaces.component.commandbutton.CommandButtonRenderer;
+import org.primefaces.util.Constants;
 
 /**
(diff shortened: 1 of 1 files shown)


melloware (approved):

melloware:
OK I tested this and the scripts are right but my original test case still doesn't work using #{component.clientId} in the JS code.

jepsar:
Ahhh.. I totally missed that part.. if there is EL in the attribute, it's only evaluated for the first button. I must dig deeper then...

PR #97, opened 2020-03-10
Fix 761 - Store and use the original ValueExpression

### LINKED ISSUE #761: CommandButtonSingleClick: Problem in Datatable

jepsar:
Fix https://github.com/primefaces-extensions/primefaces-extensions.github.com/issues/761 including EL

### DIFF
--- a/src/main/java/org/primefaces/extensions/renderer/CommandButtonSingleClickRenderer.java
+++ b/src/main/java/org/primefaces/extensions/renderer/CommandButtonSingleClickRenderer.java
@@ -42,7 +42,11 @@ public class CommandButtonSingleClickRenderer extends CommandButtonRenderer {
     }
 
     protected boolean isEligible(final CommandButton button) {
-        return button.isAjax() && button.isRendered() && !button.isDisabled() && !isConfirmation(button);
+        return button.isAjax()
+               && button.isRendered()
+               && button.getActionExpression() != null
+               && !button.isDisabled()
+               && !isConfirmation(button);
     }
 
     protected boolean isConfirmation(final CommandButton button) {
@@ -55,16 +59,8 @@ public class CommandButtonSingleClickRenderer extends CommandButtonRenderer {
     }
 
     protected String getAttributeValue(final FacesContext context, final CommandButton button, final String attribute) {
-        String key = "CommandButtonSingleClickRenderer:" + attribute;
-        ValueExpression ve = (ValueExpression) button.getAttributes().get(key);
-        if (ve == null) {
-            ve = button.getValueExpression(attribute);
-        }
-        if (ve != null) {
-            button.getAttributes().put(key, ve);
-            return (String) ve.getValue(context.getELContext());
-        }
-        return null;
+        ValueExpression ve = button.getValueExpression(attribute);
+        return ve == null ? null : (String) ve.getValue(context.getELContext());
     }
 
     protected String prefix(final String base, final String prefix) {
--- a/src/main/java/org/primefaces/extensions/renderer/CommandButtonSingleClickRenderer.java
+++ b/src/main/java/org/primefaces/extensions/renderer/CommandButtonSingleClickRenderer.java
@@ -16,10 +16,10 @@
 package org.primefaces.extensions.renderer;
 
 import java.io.IOException;
+import javax.el.ValueExpression;
 import javax.faces.context.FacesContext;
 import org.primefaces.component.commandbutton.CommandButton;
 import org.primefaces.component.commandbutton.CommandButtonRenderer;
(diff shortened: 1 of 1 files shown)


jepsar:
Wait.. I missed something else.. let me have another look at this tomorrow. I think the only thing that's currently missing is the `buildDomEvent` part.

src/main/java/org/primefaces/extensions/renderer/CommandButtonSingleClickRenderer.java
```diff
@@ -57,13 +54,17 @@ protected String getToggleJS(final String widgetVar, final boolean enabled) {
...
-        if (value != null) {
-            return value;
+    protected String getAttributeValue(final FacesContext context, final CommandButton button, final String attribute) {
+        String key = "CommandButtonSingleClickRenderer:" + attribute;
+        ValueExpression ve = (ValueExpression) button.getAttributes().get(key);
+        if (ve == null) {
+            ve = button.getValueExpression(attribute);
+        }
+        if (ve != null) {
+            button.getAttributes().put(key, ve);
```
jepsar: Don't think we need to store the expression. `button.getValueExpression(attribute)` should do.

src/main/java/org/primefaces/extensions/renderer/CommandButtonSingleClickRenderer.java
```diff
@@ -57,13 +54,17 @@ protected String getToggleJS(final String widgetVar, final boolean enabled) {
...
+            return (String) ve.getValue(context.getELContext());
```
jepsar: I think we must throw in `buildDomEvent` as used in the `commandButton` renderer.

src/main/java/org/primefaces/extensions/renderer/CommandButtonSingleClickRenderer.java
melloware: Did you test this out?  I was thinking the same thing that we need to store off the expression before it gets evaluated.

src/main/java/org/primefaces/extensions/renderer/CommandButtonSingleClickRenderer.java
melloware: Agreed.

src/main/java/org/primefaces/extensions/renderer/CommandButtonSingleClickRenderer.java
jepsar: No, didn't test this yet. I was going to.

jepsar: Yeah, we don't need to store them.

src/main/java/org/primefaces/extensions/renderer/CommandButtonSingleClickRenderer.java
jepsar: Sorry for the fuzz, it was too late to code yesterday 😄. `buildDomEvent` will be handled when we call `super.encodeMarkup(context, button);`.

src/main/java/org/primefaces/extensions/renderer/CommandButtonSingleClickRenderer.java
```diff
@@ -29,23 +29,24 @@
...
             button.setOncomplete(prefix(onComplete, getToggleJS(widgetVar, true)));
         }
         super.encodeMarkup(context, button);
     }
 
     protected boolean isEligible(final CommandButton button) {
-        return button.isAjax() && button.isRendered() && !button.isDisabled() && !isConfirmation(button);
+        return button.isAjax()
+               && button.isRendered()
+               && button.getActionExpression() != null
```
jepsar, line 47: @melloware this will prevent the issue you found earlier today.

melloware:
I am going to merge and test some more.

PR #99, opened 2020-03-24
New component: FuzzySearch #767

aripddev:
Hi @melloware ,
Below is the working example for fuzzysort component. But still need some improvement in javascript file (`primeFaces.fuzzysort.js`) to make it work different objects. It only works for sample `FuzzysortItem(String fileName, String name)` object. I put some TODO notes and I need your assistance to complete those improvements.

Bests,

### JSF

```xml
<pe:fuzzysort value="#{fuzzysortController.items}" var="item">
  <p:link outcome="#{item.fileName}" value="#{item.name}" />
</pe:fuzzysort>
```

### Controller
```java
@Named
@ViewScoped
public class FuzzysortController implements Serializable {
  
  private List<FuzzysortItem> items = new ArrayList<>();

  @PostConstruct
  public void init() {
    items.add(new FuzzysortItem("fileName1", "name1"));
    items.add(new FuzzysortItem("fileName2", "name2"));
    items.add(new FuzzysortItem("fileName3", "name3"));
  }

  public List<FuzzysortItem> getItems() {
    return items;
  }

}
```

### Model (Example)
```java
public class FuzzysortItem implements Serializable {
  private String fileName;
  
  @FuzzysortKey
  private String name;

  // constructor, getters, setters ...
}
```


src/main/resources/META-INF/resources/primefaces-extensions/fuzzysort/primefaces.fuzzysort.js
```diff
@@ -0,0 +1,52 @@
...
+
+        this.input = $(PrimeFaces.escapeClientId(this.cfg.id + "_fuzzysort-search-box"));
+        this.results = $(PrimeFaces.escapeClientId(this.cfg.id + "_fuzzysort-search-results"));
+        this.datasource = JSON.parse(this.cfg.value);
+
+        var $this = this;
+//        console.log($this);
+
+        function search() { // TODO how to move search function to search_alternative
+            $this.results.empty();
```
melloware: You should be able to move this code into search_alternative just change the $this to this when you move it.

src/main/resources/META-INF/resources/primefaces-extensions/fuzzysort/primefaces.fuzzysort.js
```diff
@@ -0,0 +1,52 @@
...
+            } else {// when there is no any input
+                $.each($this.datasource, function (index, value) {
+                    $this.results.append('<li><a href="' + value.fileName + '">' + value.name + '</a></li>'); // TODO how to get each row item format from renderChildren
+                });
+            }
+            $this.results.append('</ul>');
+        }
+
+        jQuery(document).ready(function () {
+            search();
```
melloware: Call `$this.search_alternative` here.

src/main/resources/META-INF/resources/primefaces-extensions/fuzzysort/primefaces.fuzzysort.js
```diff
@@ -0,0 +1,52 @@
...
+        });
+
+        // Run a search on input change
+        $this.input.on('input', search);
+        // Select input when escape pressed
+        document.onkeyup = (e) => { // TODO problem when using multiple implementation of fuzzysort component. So do we need to remove it?
```
melloware: So I don't think we want to attach document keyups here what is your goal with this when they hit a key anywhere on the page or do you want it only on the INPUT box?

src/main/resources/META-INF/resources/primefaces-extensions/fuzzysort/primefaces.fuzzysort.js
```diff
@@ -0,0 +1,52 @@
...
+    init: function (cfg) {
+        this.cfg = cfg;
+        this.id = this.cfg.id;
+        this.jqId = PrimeFaces.escapeClientId(this.id);
+//        this.jq = $(PrimeFaces.escapeClientId(this.cfg.target));
+
+        this.input = $(PrimeFaces.escapeClientId(this.cfg.id + "_fuzzysort-search-box"));
+        this.results = $(PrimeFaces.escapeClientId(this.cfg.id + "_fuzzysort-search-results"));
+        this.datasource = JSON.parse(this.cfg.value);
+
```
melloware: You probably want to call `PrimeFaces.skinInput(this.input);` here which styles the Input with all the right PF styling.

melloware (changes requested):

src/main/resources/META-INF/resources/primefaces-extensions/fuzzysort/primefaces.fuzzysort.js
aripdcem: Exactly, but i it is not necessary I can remove it. This is how we use it in our web page (https://aripd.com/posts/)

aripddev:
After moving search function, I get below error message while appending returned value.

```javascript
TypeError: undefined is not an object (evaluating 'this.results.append')
```

src/main/resources/META-INF/resources/primefaces-extensions/fuzzysort/primefaces.fuzzysort.js
```diff
@@ -0,0 +1,42 @@
...
+        jQuery(document).ready(function () {
+            $this.search();
+        });
+
+        // Run a search on input change
+        $this.input.on('input', $this.search);
+    },
+
+    search: function () {
+        this.results.empty();
```
melloware: First line of search should be `var $this = this;`

src/main/resources/META-INF/resources/primefaces-extensions/fuzzysort/primefaces.fuzzysort.js
```diff
@@ -0,0 +1,42 @@
...
+        this.results.append('<ul>');
+        if (this.input.val()) { // when any input entered
+            $.each(fuzzysort.go(this.input.val(), this.datasource, {keys: ['name']}), function (index, value) { // TODO how to implement search keys using a new attribute
+                this.results.append('<li><a href="' + value.obj.fileName + '">' + value.obj.name + '</a></li>'); // TODO how to get each row item format from renderChildren
```
melloware: Change `this.results` to `$this.results`

src/main/resources/META-INF/resources/primefaces-extensions/fuzzysort/primefaces.fuzzysort.js
```diff
@@ -0,0 +1,42 @@
...
+            });
+        } else {// when there is no any input
+            $.each(this.datasource, function (index, value) {
+                this.results.append('<li><a href="' + value.fileName + '">' + value.name + '</a></li>'); // TODO how to get each row item format from renderChildren
```
melloware: Change `this.results` to `$this.results`

melloware (changes requested):

src/main/resources/META-INF/resources/primefaces-extensions/fuzzysort/primefaces.fuzzysort.js
aripdcem: When I put `var $this = this;`, `$this` gives input component due to `$this.input.on('input', $this.search);`. I cannot access results and datasource variables.

melloware: Ahh because you are calling search twice.  Once with document.ready and once from the'`input` event of the `on` where this will be the InputComponent.  I htink you will need to pass in $this in your call from the input on  pass in the widget.

so `search: function (widget) {` and then change all of the this.results to `widget.results` etc..

src/main/java/org/primefaces/extensions/component/fuzzysort/FuzzysortRenderer.java
```diff
@@ -0,0 +1,136 @@
...
+        // TODO is it possible to make improvement here to have @FuzzyKey annotated class?
+        List list = (List) fuzzysort.getValue();
+        Object o = list.get(0);
+        Class<? extends Object> clazz = o.getClass();
+        Field[] fields = clazz.getDeclaredFields();
+        List<String> keys = Arrays.asList(fields).stream()
+                .filter(f -> f.getAnnotation(FuzzyKey.class) != null)
+                .map(f -> f.getName())
+                .collect(Collectors.toList());
+
```
melloware: Maybe throw a FacesException here that says `"No @FuzzyKey annotation was detected on your class " + clazz + ". Please annotate one field in the class as @FuzzyKey.`

src/main/java/org/primefaces/extensions/component/fuzzysort/FuzzysortRenderer.java
```diff
@@ -0,0 +1,136 @@
...
+        Fuzzysort fuzzysort = (Fuzzysort) component;
+
+        encodeMarkup(context, fuzzysort);
+        encodeScript(context, fuzzysort);
+    }
+
+    private void encodeScript(FacesContext context, Fuzzysort fuzzysort) throws IOException {
+        WidgetBuilder wb = getWidgetBuilder(context);
+
+        // TODO is it possible to make improvement here to have @FuzzyKey annotated class?
```
melloware: This looks OK to me as far as doing what you are trying to do.

melloware (changes requested):

src/main/java/org/primefaces/extensions/component/fuzzysort/FuzzysortRenderer.java
```diff
@@ -0,0 +1,142 @@
...
+    }
+
+    private void encodeScript(FacesContext context, Fuzzysort fuzzysort) throws IOException {
+        WidgetBuilder wb = getWidgetBuilder(context);
+
+        // TODO is it possible to make improvement here to have @FuzzysortKey annotated class?
+        List list = (List) fuzzysort.getValue();
+        Object o = list.get(0);
+        Class<? extends Object> clazz = o.getClass();
+        Field[] fields = clazz.getDeclaredFields();
```
jepsar: This will not get you fields from super classes. Is this intended?

src/main/java/org/primefaces/extensions/component/fuzzysort/FuzzysortRenderer.java
```diff
@@ -0,0 +1,142 @@
...
+        List<String> keys = Arrays.asList(fields).stream()
+                .filter(f -> f.getAnnotation(FuzzysortKey.class) != null)
+                .map(f -> f.getName())
```
jepsar: You could do `.map(Field::getName)` here

src/main/java/org/primefaces/extensions/component/fuzzysort/FuzzysortRenderer.java
```diff
@@ -0,0 +1,142 @@
...
+                .collect(Collectors.toList());
+        if (keys.isEmpty()) {
+            throw new FacesException("No @FuzzyKey annotation was detected on your class " + clazz + ". Please annotate at least one field in the class as @FuzzysortKey.");
+        }
+
+        String jsonKeys = new Gson().toJson(keys, new TypeToken<List<String>>() {
```
jepsar: @melloware was Gson already part of PFE? If not, I think it's better to switch to Json-B, as it is part of EE8.

src/main/java/org/primefaces/extensions/component/fuzzysort/FuzzysortRenderer.java
aripdcem: I try to get the annotated fields of `FuzzysortItem` model class in above example or another model object (T) provided by user. This object comes from `List<T>` which is UIData value. It works without problem, but I am not sure if there is smarter way to have it. Any performance issue etc. since I am getting the first record of the list and then extract the clazz.

src/main/resources/META-INF/resources/primefaces-extensions/fuzzysort/primefaces.fuzzysort.js
aripdcem: I have removed `document.ready` since the list has already populated in `FuzzysortRenderer`. 
I couldn't make it work search function. So the codes are working in `init` as jquery event handler.

src/main/java/org/primefaces/extensions/component/fuzzysort/FuzzysortRenderer.java
jepsar: .. it's already used by many components

aripdcem: Hi @jepsar , is it what you mean?

```java
Jsonb jsonb = JsonbBuilder.create();
//String jsonKeys = new Gson().toJson(keys, new TypeToken<List<String>>() {}.getType());
String jsonKeys = jsonb.toJson(keys, List.class);
//String jsonValue = new Gson().toJson(fuzzysort.getValue(), new TypeToken<List<Object>>() {}.getType());
String jsonValue = jsonb.toJson(fuzzysort.getValue(), List.class);
```

jepsar: @aripddev Yeah, but never mind this comment, as Gson is already used you should be able to use it as well. I already marked it as resolved.

aripddev:
Please skip the commit 27ca771
I just tried to solve issue #726 (https://github.com/primefaces-extensions/primefaces-extensions.github.com/issues/726)

melloware:
You can always submit a separate PR for the LetterAvatar fix if you fix it!  Much appreciated i had no time to spend on that one.

melloware (changes requested):
@aripddev @jepsar You are going to hate me but I think I better name for this component and all related files is "FuzzySearch" instead of "Fuzzysort" since fuzzysort is the name of the library but the action its performing is known as a FuzzySearch.   Sorry!!

aripddev:
Actually I was thinking why it is named as fuzzysort :))

aripddev:
One more improvement I need. I want to change format row item using xhtml, not from javascript. Can you guide me how to do that?

```javascript
        $this.input.on('input', function () {
            $this.results.empty();
            if ($this.input.val()) { // when any input entered
                $.each(fuzzysearch.go($this.input.val(), $this.datasource, {keys: $this.keys}), function (index, value) {
                    $this.results.append('<div><a href="' + value.obj.fileName + '">' + value.obj.name + '</a></div>'); // TODO how to get each row item format from renderChildren
                });
            } else { // when there is no input
                $.each($this.datasource, function (index, value) {
                    $this.results.append('<div><a href="' + value.fileName + '">' + value.name + '</a></div>'); // TODO how to get each row item format from renderChildren
                });
            }
        });
```

melloware:
You will have to do it with AJAX.  Look at autocomplete.js of PrimeFaces source code and look at the `search: function(query) {` method it uses Ajax to call the AutoComplete and then renders the HTML from the results.  That may be what you need to do.

Or any component that has `dynamic=true` like PrimeFaces Panel or OutputPanel.

melloware:
@aripddev just let @jepsar and I know when this PR is all ready and we can do a final review then I can merge it.

src/main/resources/META-INF/resources/primefaces-extensions/fuzzysort/primefaces.fuzzysort.js
aripdcem: created search function and finally it is working

aripddev:
I think I submitted two files by mistake, sorry.

melloware:
That is no problem I see you got the merge conflict resolved!

melloware:
@aripddev Instead of trying to replace the results with XHTML I think you should leave it in JS for speed.  You should provide two props `resultStyle` and `resultStyleClass` that can help make the DIV styling customizable.  Instead of making it an <a> link you could then have an AJAX "select" event when the user selects a result.

```xml
<p:ajax event="select" ... />
```

That way you keep the FuzzySearch fast and allow the user to take action when one is actually selected.

aripddev:
@melloware Sounds better for me too. But how to retrieve `resultStyle` and `resultStyleClass` inside javascript? I can use `FuzzySearch.ITEM_CLASS` but in this case there will be no need `resultStyle` and `resultStyleClass` and no customization by user.

melloware:
I was thinking something like this...

```javascript
var style = '';
var styleClass = '';
if (this.cfg.resultStyle) {
   style = 'style="'+this.cfg.resultStyle + '" ';
}
if (this.cfg.resultStyleClass) {
   styleClass = 'class="'+this.cfg.resultStyleClass + '" ';
}
$this.results.append('<div ' + style + styleClass + '>'+value.name+'</div>');
```

aripddev:
@melloware so the only thing remains is to implement select event. Select or click the div?

aripddev:
@melloware Am I doing right? When I clicked I can get the selected element.

melloware:
Yep but now to make it an AJAX call you need to do something like this using the PrimeFaces SelectEvent...

```javascript
        if(this.hasBehavior('select')) {
            var ext = {
                params : [
                    {name: this.id + '_itemSelect', value: itemValue}
                ]
            };

            this.callBehavior('select', ext);
        }

```

And then in your JAVA side FuzzySearchBase you need this...

```java
private static final String DEFAULT_EVENT = "select";

    private static final Map<String, Class<? extends BehaviorEvent>> BEHAVIOR_EVENT_MAPPING = MapBuilder.<String, Class<? extends BehaviorEvent>>builder()
            .put(DEFAULT_EVENT, SelectEvent.class)
            .build();

    private static final Collection<String> EVENT_NAMES = BEHAVIOR_EVENT_MAPPING.keySet();

    @Override
    public Map<String, Class<? extends BehaviorEvent>> getBehaviorEventMapping() {
        return BEHAVIOR_EVENT_MAPPING;
    }

    @Override
    public Collection<String> getEventNames() {
        return EVENT_NAMES;
    }

    @Override
    public String getDefaultEventName() {
        return DEFAULT_EVENT;
    }
```

And in your FuzzySearch...

```java
 @Override
    public void queueEvent(FacesEvent event) {
        FacesContext context = getFacesContext();
        Map<String, String> params = context.getExternalContext().getRequestParameterMap();
        String eventName = params.get(Constants.RequestParams.PARTIAL_BEHAVIOR_EVENT_PARAM);

        if (eventName != null && event instanceof AjaxBehaviorEvent) {
            AjaxBehaviorEvent ajaxBehaviorEvent = (AjaxBehaviorEvent) event;

            if (eventName.equals("select")) {
                String selectedItemValue =params.get(getClientId(context) + "_itemSelect");
                SelectEvent selectEvent = new SelectEvent(this, ajaxBehaviorEvent.getBehavior(), selectedItemValue);
                selectEvent.setPhaseId(ajaxBehaviorEvent.getPhaseId());
                super.queueEvent(selectEvent);
            }
                //e.g. blur, focus, change
                super.queueEvent(event);
            }
        }
        else {
            //e.g. valueChange, autoCompleteEven
            super.queueEvent(event);
        }
    }

```

And now you can do....
```xml
<p:ajax event="select" listener="#{controller.onSelect} />
```

```java
public void onSelect(SelectEvent event) {
    event.getValue();
}
```

melloware:
You might want to aslo add enable and disable widget methods that most widgets have like this...

```javascript
    disable: function() {
        this.input.prop('disabled', true).addClass('ui-state-disabled');
    },

    enable: function() {
        this.input.prop('disabled', false).removeClass('ui-state-disabled');
    },
```

src/main/java/org/primefaces/extensions/component/fuzzysearch/FuzzySearchRenderer.java
```diff
@@ -0,0 +1,159 @@
...
+        resultStyleClass = (resultStyleClass == null) ? FuzzySearch.ITEM_CLASS : FuzzySearch.ITEM_CLASS + " " + resultStyleClass;
+
+        writer.startElement("div", fuzzySearch);
+        writer.writeAttribute("id", clientId, null);
+        writer.writeAttribute("class", styleClass, "styleClass");
+        if (style != null) {
+            writer.writeAttribute("style", style, "style");
+        }
+        writer.writeAttribute("data-result-style", fuzzySearch.getResultStyle(), null);
+        writer.writeAttribute("data-result-style-class", fuzzySearch.getResultStyleClass(), null);
```
melloware: I think these would be better as WidgetBuilder attributes like `wb.attr('resultStyle', fuzzySearch.getResultStyle())` so they will be on the `this.cfg` object in the Javascript side.

melloware (changes requested):

src/main/resources/META-INF/resources/primefaces-extensions/letteravatar/0-letteravatar.js
```diff
@@ -5,26 +5,32 @@
         name = name || '';
         size = size || 60;
 
-        var nameSplit = String(name).toUpperCase().split(' '),
-                initials, canvas, context, dataURI;
+        var colours = [
```
jepsar, line 8: Can you merge the remote origin? These changes shouldn't be here.

src/main/resources/META-INF/resources/primefaces-extensions/sheet/1-sheet.js
```diff
@@ -521,11 +521,6 @@ PrimeFaces.widget.ExtSheet = PrimeFaces.widget.DeferredWidget.extend({
             var evt = e||window.event; // IE support
             var key = evt.charCode || evt.keyCode || 0;
             
-            // #766 do not block if just CTRL key
```
jepsar, line 524: This should also be solved with a merge.

src/main/resources/META-INF/resources/primefaces-extensions/fuzzysearch/1-fuzzysearch-widget.js
```diff
@@ -0,0 +1,73 @@
...
+    search: function (query) {
+        var $this = this;
+
+        var resultStyle = $this.jq.attr('data-result-style') || '';
+        var resultStyleClass = $this.jq.attr('data-result-style-class') || '';
+
+        var itemDisplayMarkup = '';
+        if (query) { // when any input entered
+            $.each(fuzzysearch.go(query, $this.datasource, {keys: $this.keys}), function (index, value) {
+                itemDisplayMarkup += '<div';
```
jepsar, line 71: You could do `var += ('string' + 'string' + 'string');` here

src/main/java/org/primefaces/extensions/component/fuzzysearch/FuzzySearchRenderer.java
```diff
@@ -0,0 +1,159 @@
...
+    }
+
+    private void encodeScript(FacesContext context, FuzzySearch fuzzySearch) throws IOException {
+        WidgetBuilder wb = getWidgetBuilder(context);
+
+        // TODO is it possible to make improvement here to have @FuzzySearchKey annotated class?
+        List list = (List) fuzzySearch.getValue();
+        Object o = list.get(0);
+        Class<? extends Object> clazz = o.getClass();
+        Field[] fields = clazz.getDeclaredFields();
```
jepsar: I still think you should get fields from super classes as well

src/main/resources/META-INF/resources/primefaces-extensions/fuzzysearch/1-fuzzysearch-widget.js
```diff
@@ -0,0 +1,93 @@
...
+    },
+
+    bindDynamicEvents: function () {
+        var $this = this;
+
+        this.items.on('click', function (e) {
+            if ($this.hasBehavior('select')) {
+                var ext = {
+                    params: [
+                        {name: this.id + '_itemSelect', value: this.innerHTML}
```
melloware: I think `this.id` needs to be `$this.id` does it?

melloware:
@jepsar @aripddev What if instead of FuzzyKey and the way you are loading items the UI why not use SelectItems since they are already key value pairs and that way its more like a JSF Component paradigm and people can use Converters to turn the result into an object.

Something like this...

```xml
<pe:fuzzysort value="#{fuzzysortController.valueSelected}">
     <f:selectItems value="#{fuzzysortController.items}" />
</pe:fuzzysort>
```

Where the value set to `valueSelected` is the value bound to the Input control and that way upon submission of the form it is the value selected.  And that way users can fill out selectItems any way they want.  You might even be able to exten PrimeFaces `SelectMany` and you can look at `SelectManyButtonRenderer` in PrimeFaces for example.

src/main/resources/META-INF/resources/primefaces-extensions/fuzzysearch/1-fuzzysearch-widget.js
aripdcem: Corrected to $this.id, but it does not trigger onSelect in controller.

aripdcem: $this gives component id. Shouldn't we use the selected item id?

melloware: Yeah you are right the problem is you are missing this from your Java Renderer.

```java
   @Override
    public void decode(final FacesContext context, final UIComponent component) {
        decodeBehaviors(context, component);
    }
```

aripdcem: now getting "java.lang.String cannot be cast to org.primefaces.extensions.showcase.model.fuzzysearch.FuzzySearchItem;"
When I changed `value: this.innerHTML` to `value: this`, throws NPE.
Is that the reason why you offer `f:selectItems` to turn the results into an object?

melloware: yep that is exactly why.  The more I looked at how you are doing it this is really  Select type component.

aripddev:
I am investigating `SelectMany`. Since this component will return the Object, not List, do not need to extends `SelectOne`?

jepsar:
> @jepsar @aripddev What if instead of FuzzyKey and the way you are loading items the UI why not use SelectItems since they are already key value pairs

Sounds much better to me. Also solves https://github.com/primefaces-extensions/core/pull/99#discussion_r399958493

melloware:
Yeah select one something that returns a single value.

aripddev:
Hi @melloware, I finally completed with some errors. Supposed I use `SelectEvent` but I get

```
Cannot convert [javax.faces.event.AjaxBehaviorEvent[source=org.primefaces.extensions.component.fuzzysearch.FuzzySearch@2b924cef]] of type [class javax.faces.event.AjaxBehaviorEvent] to [class org.primefaces.event.SelectEvent]
```

melloware:
Hmm look at PF code for the SelectEvent and how they use it. Its possibly I told you the wrong thing with the BEHAVIOR_NAMES when you configured it.

melloware:
In the FuzzySearchBase class.

aripddev:
Now getting NPE in controller during Ajax call

melloware:
I think you need to remove this now... as SelectOne does not have this.

```java
 @Override
    public Map<String, Class<? extends BehaviorEvent>> getBehaviorEventMapping() {
        return BEHAVIOR_EVENT_MAPPING;
    }
```

aripddev:
nope.

```
java.lang.NullPointerException at org.primefaces.extensions.showcase.controller.fuzzysearch.FuzzySearchController.valueSelectTheme(FuzzySearchController.java:42)
```

melloware:
You are missing this in FuzzySearch looks like you removed it for some reason.  But it is in SelectOne.
```java
@Override
    public void queueEvent(FacesEvent event) {
        if (event instanceof AjaxBehaviorEvent) {
            FacesContext context = getFacesContext();
            AjaxBehaviorEvent behaviorEvent = (AjaxBehaviorEvent) event;
            Map<String, String> params = context.getExternalContext().getRequestParameterMap();
            String eventName = params.get(Constants.RequestParams.PARTIAL_BEHAVIOR_EVENT_PARAM);

            if ("itemSelect".equals(eventName)) {
                Renderer renderer = ComponentUtils.getUnwrappedRenderer(
                        context,
                        "javax.faces.SelectOne",
                        "javax.faces.Menu");

                Object item = renderer.getConvertedValue(context, this, getSubmittedValue());
                SelectEvent selectEvent = new SelectEvent(this, behaviorEvent.getBehavior(), item);
                selectEvent.setPhaseId(event.getPhaseId());
                super.queueEvent(selectEvent);
            }
            else {
                super.queueEvent(event);
            }
        }
        else {
            super.queueEvent(event);
        }
    }
```

aripddev:
Now I can list and search. I can get any object from list initially rendered by `FuzzySearchRenderer`. But I cannot get any in search results. Is it because the `bindDynamicEvents` in `1-fuzzsearch-widget.js`?

src/main/resources/META-INF/resources/primefaces-extensions/fuzzysearch/1-fuzzysearch-widget.js
```diff
@@ -0,0 +1,108 @@
...
+
+    invokeItemSelectBehavior: function (event, itemValue) {
+        if (this.hasBehavior('itemSelect')) {
+            var ext = {
+                params: [
+                    {name: this.id + '_itemSelect', value: itemValue}
+                ]
+            };
+
+            this.callBehavior('select', ext);
```
melloware: This callBehavior has to be `itemSelect` now not `select`

melloware (changes requested):

melloware:
@aripddev I fear I may have steared you down the wrong path.  I would try this component with the fuzzysort.js example using dictionary.js or one of the larger ones.  Now that you are rendering SO much HTML per option with the hidden input radio etc I think this is going to generate too much HTML and bloat the browser and not perform very well.

You may ant to try and remove all the rendering of an hidden input radio per item etc.  Just use the main Input text as the value holder and when and item is selected just use the main Input box.  Because one of the problems with SelectOne is it doe not perform well with more than 100 items.  

So you might want to do some load testing.

aripddev:
@melloware no problem, I am learning the details. So I will clean all `hidden input radio`s and put it in `div` as before. Let me try.

aripddev:
I don't want to only list the label but also customize the items. To customize the item content, I am not sure how to use `p:panel` or `p:panelGrid` using `SelectOne`. It seems to me it is like `p:repeat` component.

Screen Shot 2020-04-03 at 22 24 58

melloware:
Hmmm you can do it similar to autoComplete where it has a custom mode?

https://www.primefaces.org/showcase/ui/input/autoComplete.xhtml

Click on the Custom example and type "omega".

Or you could do something where you have an attribute field called htmlTemplate="" where you pass it the raw HTML of how you want each result drawn and just replace the current DIV drawing?  Not sure the best way to do this...

melloware:
An easier solution to allow customization of the results would be to put the result rendering into its own function.
```javascript
renderItem : function(index, element) {
            itemDisplayMarkup += '<div';
            itemDisplayMarkup += (resultStyle === '' ? '' : ' style="' + resultStyle + '"');
            itemDisplayMarkup += ' class="ui-fuzzysearch-item';
            itemDisplayMarkup += (resultStyleClass === '' ? '' : ' ' + resultStyleClass) + '"';
            itemDisplayMarkup += ' data-item-value="' + element.obj.label + '">';
            itemDisplayMarkup += element.obj.label;
            itemDisplayMarkup += '</div>';
}
```

That way if you want to customize rendering anybody can just override this JS method and output any HTML they want to render the results.  To me that might be the best way to make the results customizable.

aripddev:
I am not sure. There are two types data: one is comes from fuzzysearch as array, second is the original datasource json array. The first one you will call `element.obj.label`, the second `element.value.name`. I am confused :(( I am unable to implement select call, either.

aripddev:
@melloware Do you have an idea about the reason of below error?

```
Message: 
Timestamp: 2020-04-06 17:31:32 
StackTrace: java.lang.NullPointerException
at org.primefaces.extensions.showcase.controller.fuzzysearch.FuzzySearchController.onChange(FuzzySearchController.java:38)
at org.primefaces.extensions.showcase.controller.fuzzysearch.FuzzySearchController$$OwbNormalScopeProxy0.onChange(org/primefaces/extensions/showcase/controller/fuzzysearch/FuzzySearchController.java)
at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
at java.lang.reflect.Method.invoke(Method.java:498)
at org.apache.el.parser.AstValue.invoke(AstValue.java:247)
at org.apache.el.MethodExpressionImpl.invoke(MethodExpressionImpl.java:267)
at org.apache.webbeans.el22.WrappedMethodExpression.invoke(WrappedMethodExpression.java:50)
at com.sun.faces.facelets.el.TagMethodExpression.invoke(TagMethodExpression.java:65)
at org.primefaces.behavior.ajax.AjaxBehaviorListenerImpl.processAjaxBehavior(AjaxBehaviorListenerImpl.java:71)
at javax.faces.event.AjaxBehaviorEvent.processListener(AjaxBehaviorEvent.java:100)
at javax.faces.component.behavior.BehaviorBase.broadcast(BehaviorBase.java:82)
at javax.faces.component.UIComponentBase.broadcast(UIComponentBase.java:481)
at javax.faces.component.UIViewRoot.broadcastEvents(UIViewRoot.java:847)
at javax.faces.component.UIViewRoot.processApplication(UIViewRoot.java:1396)
at com.sun.faces.lifecycle.InvokeApplicationPhase.execute(InvokeApplicationPhase.java:58)
at com.sun.faces.lifecycle.Phase.doPhase(Phase.java:76)
at com.sun.faces.lifecycle.LifecycleImpl.execute(LifecycleImpl.java:177)
at javax.faces.webapp.FacesServlet.executeLifecyle(FacesServlet.java:707)
at javax.faces.webapp.FacesServlet.service(FacesServlet.java:451)
at org.eclipse.jetty.servlet.ServletHolder.handle(ServletHolder.java:852)
at org.eclipse.jetty.servlet.ServletHandler$CachedChain.doFilter(ServletHandler.java:1604)
at org.omnifaces.filter.CacheControlFilter.doFilter(CacheControlFilter.java:239)
at org.omnifaces.filter.HttpFilter.doFilter(HttpFilter.java:108)
at org.eclipse.jetty.servlet.ServletHandler$CachedChain.doFilter(ServletHandler.java:1591)
at org.omnifaces.filter.GzipResponseFilter.doFilter(GzipResponseFilter.java:181)
at org.omnifaces.filter.HttpFilter.doFilter(HttpFilter.java:108)
at org.eclipse.jetty.servlet.ServletHandler$CachedChain.doFilter(ServletHandler.java:1591)
at org.eclipse.jetty.websocket.server.WebSocketUpgradeFilter.doFilter(WebSocketUpgradeFilter.java:226)
at org.eclipse.jetty.servlet.ServletHandler$CachedChain.doFilter(ServletHandler.java:1591)
at org.omnifaces.filter.CharacterEncodingFilter.doFilter(CharacterEncodingFilter.java:124)
at org.omnifaces.filter.HttpFilter.doFilter(HttpFilter.java:108)
at org.eclipse.jetty.servlet.ServletHandler$CachedChain.doFilter(ServletHandler.java:1583)
at org.eclipse.jetty.servlet.ServletHandler.doHandle(ServletHandler.java:542)
at org.eclipse.jetty.server.handler.ScopedHandler.handle(ScopedHandler.java:143)
at org.eclipse.jetty.security.SecurityHandler.handle(SecurityHandler.java:536)
at org.eclipse.jetty.server.handler.HandlerWrapper.handle(HandlerWrapper.java:127)
at org.eclipse.jetty.server.handler.ScopedHandler.nextHandle(ScopedHandler.java:235)
at org.eclipse.jetty.server.session.SessionHandler.doHandle(SessionHandler.java:1581)
at org.eclipse.jetty.server.handler.ScopedHandler.nextHandle(ScopedHandler.java:233)
at org.eclipse.jetty.server.handler.ContextHandler.doHandle(ContextHandler.java:1307)
at org.eclipse.jetty.server.handler.ScopedHandler.nextScope(ScopedHandler.java:188)
at org.eclipse.jetty.servlet.ServletHandler.doScope(ServletHandler.java:482)
at org.eclipse.jetty.server.session.SessionHandler.doScope(SessionHandler.java:1549)
at org.eclipse.jetty.server.handler.ScopedHandler.nextScope(ScopedHandler.java:186)
at org.eclipse.jetty.server.handler.ContextHandler.doScope(ContextHandler.java:1204)
at org.eclipse.jetty.server.handler.ScopedHandler.handle(ScopedHandler.java:141)
at org.eclipse.jetty.server.handler.ContextHandlerCollection.handle(ContextHandlerCollection.java:221)
at org.eclipse.jetty.server.handler.HandlerCollection.handle(HandlerCollection.java:146)
at org.eclipse.jetty.server.handler.HandlerWrapper.handle(HandlerWrapper.java:127)
at org.eclipse.jetty.server.Server.handle(Server.java:494)
at org.eclipse.jetty.server.HttpChannel.handle(HttpChannel.java:374)
at org.eclipse.jetty.server.HttpConnection.onFillable(HttpConnection.java:268)
at org.eclipse.jetty.io.AbstractConnection$ReadCallback.succeeded(AbstractConnection.java:311)
at org.eclipse.jetty.io.FillInterest.fillable(FillInterest.java:103)
at org.eclipse.jetty.io.ChannelEndPoint$2.run(ChannelEndPoint.java:117)
at org.eclipse.jetty.util.thread.strategy.EatWhatYouKill.runTask(EatWhatYouKill.java:336)
at org.eclipse.jetty.util.thread.strategy.EatWhatYouKill.doProduce(EatWhatYouKill.java:313)
at org.eclipse.jetty.util.thread.strategy.EatWhatYouKill.tryProduce(EatWhatYouKill.java:171)
at org.eclipse.jetty.util.thread.strategy.EatWhatYouKill.run(EatWhatYouKill.java:129)
at org.eclipse.jetty.util.thread.ReservedThreadExecutor$ReservedThread.run(ReservedThreadExecutor.java:367)
at org.eclipse.jetty.util.thread.QueuedThreadPool.runJob(QueuedThreadPool.java:782)
at org.eclipse.jetty.util.thread.QueuedThreadPool$Runner.run(QueuedThreadPool.java:918)
at java.lang.Thread.run(Thread.java:748)
```

melloware:
Hmmm no it looks like NPE trying to create your SelectEvent ?

aripddev:
Found it I think. When I remove `converter="themeConverter"`, now I get ConversionError. Why?

Since I implement fuzzysearch like selectOneButton. I try to make it work `p:selectOneButton`first with the same codebase. Then I will try to adapt it.

Screen Shot 2020-04-06 at 17 39 00

melloware:
Yep keep debugging you need the Converter to convert your id="omega" into a Theme POJO.  So something else is wrong or missing.

aripddev:
Ok got it. I tried to create `new Theme()` instead of `AvailablesThemes.getInstance().getThemes()`, that's the reason why it throws NPE. Now working. Let me adapt fuzzysearch now.

Screen Shot 2020-04-06 at 17 45 56

aripddev:
@melloware Due to possible performance issue, you told me to prefer to remove input radios and used `div` with `data-item-value` attribute to pass theme name. But I cannot have the value on controller side. Have an idea?

I tried below codes and get NPE and null respectively.
```java
Theme theme = (Theme) ((UIOutput) event.getSource()).getValue();
final FacesMessage msg = new FacesMessage(FacesMessage.SEVERITY_INFO, "Select fired: Theme is " + theme.getName(), null);
```

```java
String theme = (String) ((UIOutput) event.getSource()).getValue();
final FacesMessage msg = new FacesMessage(FacesMessage.SEVERITY_INFO, "Select fired: Theme is " + theme, null);
```

melloware:
Well for your select event that has nothing do with it right?  
JS code:
```javascript
{name: this.id + '_itemSelect', value: itemValue}
```

You are passing the selected value to the server side in the `queueEvent` method you need to simply convert that itemValue into a real object using one of the many ways to get the Converter for this compoennt.  I believe there are PF ComponetUtils that do exactly this.

aripddev:
I try to write it based on SelectOneButton, so below codes are a little different from my commits. When I complete it I will create a new PR.
 
Here is my current problem. In renderer class I implemented `getConvertedValue` like this. Since I get `null` value from `submittedValue` I am not able to make it work.
```java
@Override
public Object getConvertedValue(FacesContext context, UIComponent component, Object submittedValue) throws ConverterException {
  return ComponentUtils.getConvertedValue(context, component, (String) submittedValue);
}
```
I am sending `data-item-value` like this.
```java
String itemValueAsString = getOptionAsString(context, fuzzySearch, converter, option.getValue());
writer.writeAttribute("data-item-value", itemValueAsString, null);
```

```java
public void queueEvent(FacesEvent event) {
        if (event instanceof AjaxBehaviorEvent) {
            FacesContext context = getFacesContext();
            AjaxBehaviorEvent behaviorEvent = (AjaxBehaviorEvent) event;
            Map<String, String> params = context.getExternalContext().getRequestParameterMap();
            String eventName = params.get(Constants.RequestParams.PARTIAL_BEHAVIOR_EVENT_PARAM);

            if (DEFAULT_EVENT.equals(eventName)) {
                Object item = ComponentUtils.getConvertedValue(context, this, params.get(getClientId(context) + "_change"));
                System.out.println("item is " + item);
                System.out.println("item class is " + item.getClass());
                SelectEvent selectEvent = new SelectEvent(this, behaviorEvent.getBehavior(), item);
                selectEvent.setPhaseId(event.getPhaseId());
                super.queueEvent(selectEvent);
            }
            else {
                super.queueEvent(event);
            }
        }
        else {
            super.queueEvent(event);
        }
    }
```

```javascript
triggerChange: function (button) {
        var itemValue = button.attr('data-item-value');
        console.log(itemValue);

        if (this.cfg.change) {
            this.cfg.change.call(this);
        }

        console.log(itemValue);
        var ext = {
            params: [
                {name: this.id + '_change', value: itemValue}
            ]
        };

        this.callBehavior('change', ext);
    },
```

jepsar:
Closing this as there is a new PR

PR #101, opened 2020-04-03
LetterAvatar: Using HTML + CSS to generate it #774

aripddev:


aripddev:
used style inside div and span to generate letter avatar and removed .css and .js files

melloware:
Looks like you have some conflicts.

Also can you look at this ticket: https://github.com/primefaces-extensions/primefaces-extensions.github.com/issues/695

Three is a reproducer ZIP file in that ticket please make sure that still works.  That is why I had to add the widget JS to Letter Avatar.

aripddev:
@melloware I have checked the ticket, there is no problem in datable, works well with PF 8.0 and PFE 8.0.1-SNAPSHOT.
Do I need to keep the .js file because of conflict?

aripddev:
@melloware One more thing. How can I adjust text color depending on background color? Currently it is white and it can be changed by user which is not proper for some backgrounds. For example white text color is not good for yellow background? Do we make it dynamic depends on background color as @jepsar suggested hash+hue algorithm before?

melloware:
with pure CSS not sure you can adjust per the background color.  Maybe @jepsar knows a way?  

I am not sure what it's saying the conflict is but until you resolve it it won't let me merge this PR?

melloware:
I am going to merge this PR if its OK with you. If you figure out how to contrast the background color you can always submit a future PR.

src/main/java/org/primefaces/extensions/component/letteravatar/LetterAvatar.java
```diff
@@ -53,8 +51,8 @@
         styleClass,
         value,
         rounded,
-        width,
-        height;
+        size,
+        color;
```
jepsar, line 55: Wouldn't it be better to call this `textColor`?

src/main/java/org/primefaces/extensions/component/letteravatar/LetterAvatarRenderer.java
```diff
@@ -43,23 +47,43 @@ public void encodeEnd(final FacesContext context, final UIComponent component) t
...
+//                .collect(Collectors.joining(""));
+//              or .toArray(String[]::new);
+        initials = initials.length() == 1 ? initials : initials.charAt(0) + initials.substring(initials.length() - 1);
+
         final Boolean rounded = letterAvatar.isRounded();
 
         final String clientId = letterAvatar.getClientId(context);
-        final String style = letterAvatar.getStyle();
+
+        String backgroundColor = "hsl(" + hue(value) + ", 100%, 50%)";
```
jepsar, line 79: If you would allow the user to set the saturation and lightness, that would solve the possible conflict with the text color.

src/main/java/org/primefaces/extensions/component/letteravatar/LetterAvatarRenderer.java
```diff
@@ -71,18 +95,56 @@ public void encodeMarkup(FacesContext context, LetterAvatar letterAvatar) throws
...
+        int result = 0;
+        for (int i = 0; i < str.length(); i++) {
+            result = ((result << 5) - result) + str.charAt(i);
+            result |= 0;
+        }
+        return result;
+    }
+
+    private int hue(String str) {
+        return Math.abs(hash(str) % 360);
```
jepsar, line 119: Don't we need some modulo here to prevent colors that are too close to other colors?

src/main/java/org/primefaces/extensions/component/letteravatar/LetterAvatarRenderer.java
```diff
@@ -71,18 +95,56 @@ public void encodeMarkup(FacesContext context, LetterAvatar letterAvatar) throws
...
+        Map<String, String> map = new LinkedHashMap<>();
+        map.put("color", color);
+        map.put("background-color", backgroundColor);
+        if (rounded) {
+            map.put("border-radius", "50%");
+        }
+        map.put("height", size);
+        map.put("text-align", "center");
+        map.put("width", size);
+        return map.entrySet()
```
jepsar, line 132: This could be DRYed by adding a method

src/main/java/org/primefaces/extensions/component/letteravatar/LetterAvatarRenderer.java
aripdcem: I tried to convert your algorithm from javascript to java :) is it wrong?

jepsar: I'm not saying it is wrong. I'm suggesting an improvement.

src/main/java/org/primefaces/extensions/component/letteravatar/LetterAvatarRenderer.java
melloware: @jepsar its all checked into master now if you want to make any tweaks!

jepsar: Alright. I'll have a look.

PR #111, opened 2020-05-19
Added Content component

jepsar:
First PR for https://github.com/primefaces-extensions/primefaces-extensions.github.com/issues/797

Showcase will follow.

### DIFF
--- a/src/main/java/org/primefaces/extensions/component/content/Content.java
+++ b/src/main/java/org/primefaces/extensions/component/content/Content.java
@@ -1,0 +1,107 @@
+/*
+ * Copyright 2011-2020 PrimeFaces Extensions
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.primefaces.extensions.component.content;
+
+import java.util.Locale;
+import javax.faces.component.UIComponentBase;
+import javax.faces.context.FacesContext;
+import org.primefaces.util.LocaleUtils;
+
+/**
+ * <code>Content</code> component.
+ *
+ * @author Jasper de Vries &lt;jepsar@gmail.com&gt;
+ * @since 8.0.3
+ */
+public class Content extends UIComponentBase {
+
+    public static final String COMPONENT_TYPE = "org.primefaces.extensions.component.Content";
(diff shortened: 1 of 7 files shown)


src/main/java/org/primefaces/extensions/component/content/ContentRenderer.java
```diff
@@ -0,0 +1,113 @@
...
+
+    protected void encodeMarkup(final FacesContext context, final Content content) throws IOException {
+        String value = Files.readString(resolve(context, content));
+        if (content.isEvalEl()) {
+            value = evaluateEl(context, value);
+        }
+        if (content.isEscape()) {
+            value = value
+                        .replace("<", "&lt;")
+                        .replace(">", "&gt;");
```
melloware: I think you can use EscapeUtils.escapeXML() which should be a PrimeFaces class.

src/main/java/org/primefaces/extensions/config/PrimeExtensionsEnvironment.java
```diff
@@ -59,6 +61,10 @@ public static void setCurrentInstance(PrimeExtensionsEnvironment pfeEnv, FacesCo
         }
     }
 
+    public boolean isCommonmarkAvailable() {
```
melloware, line 64: Maybe just call this `isMarkdownAvailable`

src/main/java/org/primefaces/extensions/component/content/Content.java
```diff
@@ -0,0 +1,107 @@
...
+import org.primefaces.util.LocaleUtils;
+
+/**
+ * <code>Content</code> component.
+ *
+ * @author Jasper de Vries &lt;jepsar@gmail.com&gt;
+ * @since 8.0.3
+ */
+public class Content extends UIComponentBase {
+
```
melloware, line 30: I like everything about this except i am not sure of the name `Content` as too generic.  Maybe `StaticContent`??? @Rapster what are your thoughts?

melloware (changes requested):

src/main/java/org/primefaces/extensions/config/PrimeExtensionsEnvironment.java
jepsar: I opted to use the library name.

src/main/java/org/primefaces/extensions/component/content/Content.java
melloware: Instead of `Content` maybe `LocalizedContent`?

Rapster: I'm not sure I understand exactly what the component does... By the sound of this:
> We have several applications where we need to output larger blocks of HTML text which can differ per context / country, which we have multiple languages of and can fallback to a default context and language. Message bundles just was not the tool for the job. As I could not find any existing tools I decided to create a component.

It sounds like a specific component is not necessary, but again I don't fully understand the context, and how markdown comes into the picture? Maybe if you could provide a jar for the showcase so we can have a quick look, that would be nice :wink:

tandraschko:
Currently also a -0.5 from my side.
Looks like it can be done with a simple controller (with only some lines of code) in your webapplication + h:outputText escape=false.

PR #115, opened 2020-08-23
Fix #815: CodeScanner: add start, stop

### LINKED ISSUE #141: CodeScanner: add start, stop functionality
Add (auto)start attribute and start and stop functions to widget

jepsar:
Fix https://github.com/primefaces-extensions/primefaces-extensions.github.com/issues/815

### DIFF
--- a/src/main/java/org/primefaces/extensions/component/codescanner/CodeScanner.java
+++ b/src/main/java/org/primefaces/extensions/component/codescanner/CodeScanner.java
@@ -61,7 +61,7 @@ public class CodeScanner extends UIComponentBase implements Widget, ClientBehavi
         styleClass,
         width,
         height,
-        start
+        autoStart
     }
 
     public enum ReaderType {
@@ -137,12 +137,12 @@ public class CodeScanner extends UIComponentBase implements Widget, ClientBehavi
         getStateHelper().put(PropertyKeys.height, height);
     }
 
-    public boolean getStart() {
-        return (Boolean) getStateHelper().eval(PropertyKeys.start, true);
+    public boolean getAutoStart() {
+        return (Boolean) getStateHelper().eval(PropertyKeys.autoStart, true);
     }
 
-    public void setStart(boolean start) {
-        getStateHelper().put(PropertyKeys.start, start);
+    public void setAutoStart(boolean autoStart) {
+        getStateHelper().put(PropertyKeys.autoStart, autoStart);
     }
 
     @Override
--- a/src/main/java/org/primefaces/extensions/component/codescanner/CodeScanner.java
+++ b/src/main/java/org/primefaces/extensions/component/codescanner/CodeScanner.java
@@ -60,7 +60,8 @@ public class CodeScanner extends UIComponentBase implements Widget, ClientBehavi
         style,
         styleClass,
         width,
-        height
+        height,
+        start
     }
 
     public enum ReaderType {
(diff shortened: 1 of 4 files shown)


src/main/java/org/primefaces/extensions/component/codescanner/CodeScanner.java
```diff
@@ -60,7 +60,8 @@
         style,
         styleClass,
         width,
-        height
+        height,
+        start
```
melloware: Can you check what poller calls it?  I think it may be "autoStart"

src/main/resources/META-INF/resources/primefaces-extensions/codescanner/codescanner-widget.js
```diff
@@ -36,12 +45,7 @@ PrimeFaces.widget.ExtCodeScanner = PrimeFaces.widget.BaseWidget.extend({
...
             }
-
         }
-
-        var $this = this;
-        this.video = $('video', this.jq)[0];
-        this.codeReader = getReader(cfg.type);
         this.codeReader.listVideoInputDevices()
                 .then((videoInputDevices) => {
                     var deviceId = videoInputDevices[0].deviceId;
```
melloware, line 51: Do not use lambdas.  PF supports ES5 only.

melloware (changes requested):

src/main/resources/META-INF/resources/primefaces-extensions/codescanner/codescanner-widget.js
jepsar: Ah.. Zxing is ES6.. so we cannot use that?

src/main/java/org/primefaces/extensions/component/codescanner/CodeScanner.java
jepsar: Yes, it is. I'll change it.

src/main/resources/META-INF/resources/primefaces-extensions/codescanner/codescanner-widget.js
jepsar: .. I cannot find any info on what version PF is using... Isn't mentioning the minimum ES version in the documentation good enough in order to be able to use the library?

melloware: Well I just know from experience and previous commits. can you try and run `mvn clean package -Pcompress` and see if it fails

melloware:
I bumped the pom.xml from ECMASCRIPT5 to ECMASCRIPT_2016 and it seems to be compressing now!  Let me test the showcase make sure nothing is broken.

melloware:
Everything looking good so do an update and then a `mvn clean install -Pcompress` to test.

melloware:
OK nope its breaking Layout and other components compiling for ECMASCRIPT_2016.  Can you just rewrite the one section:
```javascript
 this.codeReader.listVideoInputDevices()
                .then((videoInputDevices) => {
                    var deviceId = videoInputDevices[0].deviceId;
                    $this.codeReader.decodeFromVideoDevice(deviceId, $this.video, (result, err) => {
                        if (result) {
                            handleResult(result);
                        }
                        if (err && !(err instanceof ZXing.NotFoundException)) {
                            console.error(err);
                        }
                    })
                })
                .catch((err) => {
                    console.error(err)
                });
```

I think that is the only thing failing.  Make sure to test by building with `-Pcompress` and testing the showcase.

jepsar:
Yeah, I'll rewrite it.

melloware:
No worries I just checked in a fix.  I turned compression OFF for the codescanner so it just gets combined not compressed and its working.
```xml
                            <aggregations>
                                <aggregation>
                                    <inputDir>${resources.dir.compressed}</inputDir>
                                    <outputFile>${resources.dir.compressed}/codescanner/codescanner.js</outputFile>
                                    <withoutCompress>true</withoutCompress>
                                </aggregation>
                            </aggregations>
```
This fix seems to work for now.

PR #116, opened 2020-08-24
Fix #816: CodeScanner: Added onsuccess and onerror client side listeners

jepsar:
https://github.com/primefaces-extensions/primefaces-extensions.github.com/issues/816

### DIFF
--- a/src/main/resources/META-INF/resources/primefaces-extensions/codescanner/codescanner-widget.js
+++ b/src/main/resources/META-INF/resources/primefaces-extensions/codescanner/codescanner-widget.js
@@ -32,35 +32,38 @@ PrimeFaces.widget.ExtCodeScanner = PrimeFaces.widget.BaseWidget.extend({
 
     start: function() {
         var $this = this;
-        function handleResult(result) {
-            if ($this.hasBehavior('codeScanned')) {
-                var ext = {
-                    params: [{
-                            name: $this.id + '_value',
-                            value: result.text
-                        }, {
-                            name: $this.id + '_format',
-                            value: result.format
-                        }]
-                };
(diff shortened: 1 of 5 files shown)


src/main/resources/META-INF/resources/primefaces-extensions/codescanner/codescanner-widget.js
```diff
@@ -32,35 +32,38 @@ PrimeFaces.widget.ExtCodeScanner = PrimeFaces.widget.BaseWidget.extend({
...
+                                value: result.text
+                            }, {
+                                name: $this.id + '_format',
+                                value: result.format
+                            }]
+                    };
+                    $this.callBehavior('codeScanned', ext);
+                }
+                if ($this.cfg.onsuccess) {
+                    eval($this.cfg.onsuccess);
```
melloware: Don't use `eval` it violates CSP.  I think you want to do a `call` right?

melloware (changes requested):

src/main/resources/META-INF/resources/primefaces-extensions/codescanner/codescanner-widget.js
jepsar: Thanks for the pointer.. I was not sure how to do this. I think it's okay right now.

melloware (approved):

src/main/resources/META-INF/resources/primefaces-extensions/codescanner/codescanner-widget.js
melloware: Yep that will do it!

PR #121, opened 2020-08-26
Fix #742: TimeAgo: Use zone from value if possible

### LINKED ISSUE #148: TimeAgo should be able to render the title in the original time zone
The should be an attribute to either use the system zone or the zone present in the time object

jepsar:
Fix https://github.com/primefaces-extensions/primefaces-extensions.github.com/issues/742


melloware (approved):
Excellent

PR #149, opened 2020-08-27
Added maven.config

jepsar:
This allows you to only having to change the version at a single location

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -38,3 +38,5 @@
 # virtual machine crash logs, see http://www.java.com/en/download/help/error_hotspot.xml
 hs_err_pid*
 target/
+
+**/.flattened-pom.xml
--- a/.mvn/maven.config
+++ b/.mvn/maven.config
@@ -1,0 +1,1 @@
+-Drevision=9.0-SNAPSHOT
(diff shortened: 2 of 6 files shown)


melloware:
nice but this does not affect the Deploy plugin right?

tandraschko:
i dont really like it
never seen a project which uses the pattern (PF or many Apache projects)

tandraschko:
and you can change all versions via single command: mvn versions:set

melloware:
Here is what I do for PrimeSelenium...
```
Update pom.xml to new version e.g 8.0.2
Run mvn -N versions:update-child-modules -DgenerateBackupPoms=false to update all child modules versions
Commit and push the changes to GitHub
In GitHub create a new Release titled 8.0.2 to tag this release
Run mvn clean deploy -Prelease to push to Maven Central
```

tandraschko:
mvn versions:set -DnewVersion=10.0-SNAPSHOT 

will change it in the root pom and update all submodules

melloware:
Awesome let me go update my notes.  I have always done the update-child-modules but i like yours better.

jepsar:
> and you can change all versions via single command: mvn versions:set

Ha, didn't know that. Cool!

melloware:
I updated the Release instructions on the README.MD of this project.

PR #157, opened 2020-09-12
Fix #154: CommandButtonSingleClickRenderer prevent widget not available

### LINKED ISSUE #154: CommandButtonSingleClick: widget is not available if update results in the button not being rendered
JS console will display:

> Widget for var 'widget_xxx' not available!

It doesn't really break anything, but we should check for it.

jepsar:


melloware (approved):
I like it so it skips that PrimeFaces debug output and since we are already checking its existence!

PR #159, opened 2020-09-23
Fix #158: Add BreadCrumbAccessibleRenderer

### LINKED ISSUE #158: New renderer: accessible BreadCrumb
The current PF breadcrumb renderer does not follow the ARIA best practices, see:

https://github.com/primefaces/primefaces/issues/4928

With a custom renderer we could improve that. Only downside is, because of the different HTML, you would need to do the styling, as it differs per theme.

jepsar:
Fix #158

### DIFF
--- a/core/src/main/java/org/primefaces/extensions/component/codescanner/CodeScanner.java
+++ b/core/src/main/java/org/primefaces/extensions/component/codescanner/CodeScanner.java
@@ -70,7 +70,7 @@ public class CodeScanner extends UIComponentBase implements Widget, ClientBehavi
         deviceId,
         forVal("for");
 
-        String toString;
+        private String toString;
         PropertyKeys(String toString) {
             this.toString = toString;
         }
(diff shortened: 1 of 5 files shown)


melloware:
hmmm any thoughts to fixing this in PF core?

jepsar:
> hmmm any thoughts to fixing this in PF core?

You will keep the issue of styling until something is done to fix (all) the themes. But, you could add a flag to trigger the accessible version and warn the user to fix the styling.. not ideal, but it's something.

melloware:
Yes I think we should fix this in PF core. Lets just fix it and we won't merge until Mert updates the CSS in the themes.  And that is only in the new themes we no longer support the old Jquery ThemeRoller themes anyway.

jepsar:
So, what do you suggest? Add a flag to toggle between regular and accessible? Or just replace it with the accessible variant (requiring the themes to be updated)

melloware:
Let's add an "accessible "flag that is false by default. That will also make it backward compatible with older themes. Once all the modern themes are updated we can flip it to be True by default. But that way we fix the accessibility issues in core.  Sound good?

jepsar:
I'll open a PR at PF and close this

melloware:
Thank you!

PR #162, opened 2020-09-30
LetterAvatar performance improvements

jepsar:


tandraschko:
i would not use a StringBuilder here, normally JVM does StringBuilder conversation itself

tandraschko:
in theory our SharedStringBuilder would be a performance improvement here as not always a new StringBuilder will be initialized.
However this makes the code unreadable. If you check PF, we only use it in very core stuff.

melloware (approved):

PR #163, opened 2020-10-01
Sonar @SuppressWarnings("java:S115")

jepsar:
@melloware I would like to get rid of SonarQube issues with high severity. What's your opinion on suppressing issues that we cannot fix?

### DIFF
--- a/core/src/main/java/org/primefaces/extensions/component/calculator/Calculator.java
+++ b/core/src/main/java/org/primefaces/extensions/component/calculator/Calculator.java
@@ -60,6 +60,7 @@ public class Calculator extends UIComponentBase implements ClientBehaviorHolder,
 
     private Locale appropriateLocale;
 
+    @SuppressWarnings("java:S115")
     protected enum PropertyKeys {
 
         // @formatter:off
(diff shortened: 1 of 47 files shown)


melloware:
I have been doing that supressing things like where it doesn't like PropertyKeys enumeration is in UPPERCASE etc.  So feel free!

core/src/main/java/org/primefaces/extensions/component/calculator/Calculator.java
```diff
@@ -60,6 +60,7 @@
 
     private Locale appropriateLocale;
 
+    @SuppressWarnings("java:S115")
```
melloware, line 63: Oh cool, we probably need this supresswarning on every single PropertyKeys.

jepsar:
@melloware there are a few complexity issues which I didn't want to solve here. I think we need PRs per class. Especially the exporter is horrible.

melloware:
Agrees the Exporter is in really bad shape. I cleaned it up a few years ago but only really glaring things. I originally wanted to delete the component since PF version has almost the same features now but I think people still use this one. Good luck!

melloware (approved):

PR #165, opened 2020-10-03
Sonar: ExtLangUtils

jepsar:


jepsar:
Sonar was also complaining about duplicates in `unescapeXml`. I can modify that code and use a map. That will also bring down the complexity.

core/src/main/java/org/primefaces/extensions/util/ExtLangUtils.java
```diff
@@ -91,9 +91,7 @@ public static String deleteWhitespace(final String str) {
     }
 
     public static String[] subarray(final String[] array, int startIndexInclusive, int endIndexExclusive) {
-        if (array == null) {
-            return null;
-        }
+        Objects.requireNonNull(array);
```
melloware: THis throws an exception instead of gracefully returning NULL like it was before?

melloware (changes requested):

core/src/main/java/org/primefaces/extensions/util/ExtLangUtils.java
jepsar: Returning `null` is not considered graceful, but a code smell instead. See https://sonarcloud.io/project/issues?id=org.primefaces.extensions%3Aprimefaces-extensions-parent&issues=AXRFAP8Ks0QFIU1_liC6&open=AXRFAP8Ks0QFIU1_liC6

Changed it to an empty array as suggested.

Having the input array as a required argument seemed better to me.

By the way, it is only used here:

https://github.com/primefaces-extensions/primefaces-extensions/pull/165/files#diff-ced6042d219d333f17bb9604cc8db9c4R108

PR #168, opened 2020-10-04
Sonar: RemoteCommand, SlideOut and TimePicker

jepsar:


tandraschko:
-1 from my side to use constants for the all the Strings
Makes it more unreadable, better to implement a typesafe builder in the future

And for looping children, use getChildCount, this can avoid child list instances

jepsar:
Alright, let me close this PR then and revert https://github.com/primefaces-extensions/primefaces-extensions/pull/164

PR #254, opened 2020-12-26
Fix #253: Timer locale support

### LINKED ISSUE #253: Timer: Locale support
**Is your feature request related to a problem? Please describe.**
There is no i18n support for timer component. For `human` format, it is shown only English time words, such as `an hour`, `a few minutes` so on.

**Describe the solution you'd like**
Need to implement the solution mentioned moments documentation at https://momentjs.com/docs/#/i18n/

aripddev:
I think it is done

### DIFF
--- a/core/src/main/java/org/primefaces/extensions/component/timer/Timer.java
+++ b/core/src/main/java/org/primefaces/extensions/component/timer/Timer.java
@@ -15,6 +15,7 @@
  */
 package org.primefaces.extensions.component.timer;
 
+import java.util.Locale;
 import javax.el.MethodExpression;
 import javax.faces.application.ResourceDependency;
 import javax.faces.component.UIComponentBase;
@@ -23,6 +24,7 @@ import javax.faces.context.FacesContext;
 import org.primefaces.component.api.AjaxSource;
 import org.primefaces.component.api.Widget;
 import org.primefaces.util.Constants;
+import org.primefaces.util.LocaleUtils;
 
 /**
  * Timer component
@@ -44,6 +46,8 @@ public class Timer extends UIComponentBase implements Widget, AjaxSource {
     public static final String STYLE_CLASS = "ui-timer ui-widget ui-widget-header ui-corner-all";
     private static final int DEFAULT_TIMEOUT = 10;
     private static final int DEFAULT_INTERVAL_MS = 1000;
+
+    private Locale appropriateLocale;
 
     @SuppressWarnings("java:S115")
     protected enum PropertyKeys {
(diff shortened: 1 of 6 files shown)


core/src/main/resources/META-INF/resources/primefaces-extensions/timer/timer.js
```diff
@@ -49,7 +50,7 @@ PrimeFaces.widget.ExtTimer = PrimeFaces.widget.BaseWidget.extend({
             if ("percentage" === format) {
                 value = ((this.currentTimeout * 100) / this.originalTimeout) + "%";
             } else if ("human" === format) {
-                value = moment.duration(this.currentTimeout, 'seconds').humanize();
+                value = moment.duration(this.currentTimeout, 'seconds').locale(this.locale).humanize();
             } else {
                 value = moment.utc(moment.duration(this.currentTimeout, 'seconds').asMilliseconds()).format(format);
```
melloware: Should this moment value here also use locale?

core/src/main/java/org/primefaces/extensions/component/timer/Timer.java
```diff
@@ -345,6 +346,14 @@ public void setForward(final boolean forward) {
         getStateHelper().put(PropertyKeys.forward, forward);
     }
```
melloware, line 361: Also look at how Calculator does `locale`...
```java
public Locale calculateLocale() {
        if (appropriateLocale == null) {
            final FacesContext fc = FacesContext.getCurrentInstance();
            appropriateLocale = LocaleUtils.resolveLocale(fc, getLocale(), getClientId(fc));
        }
        return appropriateLocale;
    }
```
LocaleUtils is smart and will use Locale if defined or the global faces-config.xml locale or the browsers locale.

core/src/main/java/org/primefaces/extensions/component/timer/Timer.java
```diff
@@ -345,6 +346,14 @@ public void setForward(final boolean forward) {
...
 
+    public java.lang.String getLocale() {
+        return (java.lang.String) getStateHelper().eval(PropertyKeys.locale, "en");
+    }
```
melloware, line 364: Default value should be NULL instead of "en" and the above CalculateLocale will figure out the default.

melloware (changes requested):

core/src/main/resources/META-INF/resources/primefaces-extensions/timer/timer.js
aripdcem: Right, I will add the same for custom format

melloware (approved):

melloware:
Thank for the PR!  Great work.

PR #265, opened 2021-01-05
#257 FAB: Add PF badge support

### LINKED ISSUE #257: FAB: Support PF Badge
**Is your feature request related to a problem? Please describe.**
@jepsar since PE Badge works with FAB we should make it work with the new PrimeFaces Badge as that would be the only functionality lost if we deprecated and removed PE Badge.

jepsar:
Fix #257


melloware (approved):
Nice work!

PR #297, opened 2021-01-30
Fix #296 Added LightSwitch

### LINKED ISSUE #296: New component: LightSwitch
Automatic dark or light theme selection based on the OS settings.

```xml
<pe:lightSwitch selected="#{userSettings.currentThemeName}" light="saga" dark="arya"/>
```

Declare which theme is light and which is dark and your bound EL bean for where you are storing the value

jepsar:
Fix #296

Showcase is still to do.. I'll take care of that after this has been reviewed

### DIFF
--- a/core/src/main/java/org/primefaces/extensions/component/lightswitch/LightSwitchRenderer.java
+++ b/core/src/main/java/org/primefaces/extensions/component/lightswitch/LightSwitchRenderer.java
@@ -55,6 +55,7 @@ public class LightSwitchRenderer extends CoreRenderer {
                     .attr("selected", lightSwitch.getSelected())
                     .attr("light", lightSwitch.getLight())
                     .attr("dark", lightSwitch.getDark())
+                    .attr("automatic", lightSwitch.isAutomatic())
                     .finish();
     }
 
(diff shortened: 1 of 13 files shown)


melloware (commented):
Cool let me check this out!

melloware (approved):
Looks good to me.

melloware:
I also just fixed some oddities with our themeswitcher and deleted a bunch of old JS methods for theme switching in core.js that weren't even being used!

core/src/main/java/org/primefaces/extensions/component/lightswitch/LightSwitchRenderer.java
```diff
@@ -0,0 +1,63 @@
...
+
+/**
+ * Renderer for the {@link LightSwitch} component.
+ *
+ * @author Jasper de Vries &lt;jepsar@gmail.com&gt;
+ * @since 10.0
+ */
+public class LightSwitchRenderer extends CoreRenderer {
+
+    public static final String ATTR_THEME = "org.primefaces.extensions.component.lightswitch.THEME";
```
jepsar: Remove

melloware (changes requested):
I am thinking one more good change might be a `disabled` flag on the component and two new widget methods for `enable` and `disable`.  This way people can control it if they need to.

jepsar:
> I am thinking one more good change might be a `disabled` flag on the component and two new widget methods for `enable` and `disable`. This way people can control it if they need to.

I added an `automatic` attribute instead and some API functions to control the switch. Disabling does not make sense to me as it's only executed on page load. If you don't want it to run you can now use `automatic="false"` or `rendered="false"`.

melloware (approved):
Ok good you are safe to squash and merge it!

PR #304, opened 2021-02-03
Fix #303: LightSwitch switch event

### LINKED ISSUE #303: LightSwitch: Add switch event
.. so you can update components on switch. For example, on the showcase the theme selector is showing an outdated theme on switch.

jepsar:
Fix #303

This adds the event, but does not (yet) fix the bug we've found with the non-event variant.

### DIFF
--- a/core/src/main/java/org/primefaces/extensions/component/codescanner/CodeScanner.java
+++ b/core/src/main/java/org/primefaces/extensions/component/codescanner/CodeScanner.java
@@ -218,7 +218,7 @@ public class CodeScanner extends UIComponentBase implements Widget, ClientBehavi
 
     @Override
     public Collection<String> getUnobstrusiveEventNames() {
-        return EVENT_NAMES;
+        return getEventNames();
     }
 
     @Override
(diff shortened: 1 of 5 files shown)


jepsar:
Not perfect, but it's working. Can be tested by removing `<p:ajax update="themeSelectMenu"/>`.

melloware (approved):
Nice changes.  I figured it had to be in its own form that is a good idea.

melloware:
Working well for me.

jepsar:
> Working well for me.

Did you also check without `<p:ajax update="themeSelectMenu"/>`?

jepsar:
To properly fix this, I think we need to be able to bypass

https://github.com/primefaces/primefaces/blob/47a1f00dc3f2be3a66e37e75330eba65a9e380d8/src/main/java/org/primefaces/context/PrimePartialResponseWriter.java#L268-L295

melloware:
I did not try it without the AJAX.   Let me investigate today and see if I come to the same conclusion.

PR #316, opened 2021-02-21
[MonacoEditor] feat: new component: MonacoEditor

### LINKED ISSUE #305: New component: Monaco editor
The [Monaco Editor](https://microsoft.github.io/monaco-editor/index.html) is the code editor which powers VS Code, with the features better described here.

Two years, ago I started making a JSF wrapper for the monaco editor (https://github.com/blutorange/primefaces-monaco) so we could use it in our JSF pages. It's already implemented as a PrimeFaces widget. By now the wrapper component works well enough that I was wondering if we could add it to PrimeFaces extensions ? I would need to make some adjustments, but it should not be too hard.

* Allows a language / locale to be specified.
* There's an inline version and an iframe version. The iframe version is useful when you want to use a different language service configuration.
* There's an extender that lets you setup the monaco editor via custom script.
* The github project includes a demo JSF application that I also use for testing, see https://github.com/blutorange/primefaces-monaco#demo if you want to try it out.

blutorange:
It's done for now, hope it can get into RC2 : ) #305 

* New component: Monaco Editor with an inline an iframe variant
* Separate submodule for the monaco editor webpack build
* Monaco editor requires Promise to be available. Microsoft does officially not support IE anymore with Monaco Editor, all other browsers support Promises
* I changed the input language for the optimizer (closure compiler) to ECMA_18, this lets us use a few more syntax features that are transpiled away, the target is still at ES5
* Added some examples to the showcase that illustrate how to use the component. This is the commit for RC2, I'd like to add additional examples to the showcase later
* Suppport localization, uses default locale or user-specified, supports custom localization (see showcase).


melloware:
I will review in the morning and merge if it looks good. you rock!

PR #334, opened 2021-03-14
Fix 326: CodeScanner: should start with back side camera by default

### LINKED ISSUE #326: CodeScanner: should start with back side camera by default

jepsar:
Fix #326: CodeScanner: should start with back side camera by default


melloware:
Nice this was a good fix

PR #558, opened 2021-09-23
[MonacoEditor] Remove outdated dependency replace-in-file

blutorange:
This removes the `replace-in-file` dependency used for building the monaco editor bundle. It pulled in `ansi-regex` as a transitive dependency which has a moderate severity CVE in version <=5.0.0. Since (a) `replace-in-file` didn't get an update in the last few months and (b) it was only used in one place in the build script to replace some data in a file, I removed the dependency and just replace the data manually now.


melloware (approved):

melloware:
Thanks for fixing!

PR #641, opened 2021-12-06
PFE#640 - initial commit

### LINKED ISSUE #640: New component: Keynote
I am thinking of adding HTML presentation component for JSF.
I'm not sure if we should use the name "keynote" or the original name "revealjs".
reveal.js is an open source HTML presentation framework. It's a tool that enables anyone with a web browser to create fully-featured and beautiful presentations. It can be accessed from https://revealjs.com.

```xhtml
<pe:keynote>
  <pe:keynoteItem>Slide 1</pe:keynoteItem>
  <pe:keynoteItem>Slide 2</pe:keynoteItem>
  <pe:keynoteItem>Slide 3</pe:keynoteItem>
</pe:keynote>
```

Renders
```html
<div class="reveal">
  <div class="slides">
    <section>Slide 1</section>
    <section>Slide 2</section>
    <section>Slide 3</section>
  </div>
</div>
```

aripddev:
Keynote component with basic features without theming. I am unable to implement theming.

I think below code in pom.xml is merging all theme files into one. Since each theme is available as a separate stylesheet, it doesn't work. I can get theme name but I don't know how to load relevant css file.

```xml
<include>keynote/theme/*.css</include>
```


aripddev:
I'm making progress. I have found it how to add a resource inside `DocumentViewerRenderer`.

core/pom.xml
```diff
@@ -192,6 +195,32 @@
...
+                                    <inputDir>${resources.dir.compressed}</inputDir>
+                                    <outputFile>${resources.dir.compressed}/keynote/keynote.js</outputFile>
+                                    <withoutCompress>${withoutCompress}</withoutCompress>
+                                </aggregation>
+                            </aggregations>
+                        </resourcesSet>
+                        <resourcesSet>
+                            <includes>
+                                <include>keynote/lib/**/*.css</include>
+                                <include>keynote/theme/*.css</include>
```
melloware: Won't this accidentally combine all the themes into one giant CSS file?  If this is correct that is OK but makre sure you test switching themes after building with `mvn clean install -Pcompress`

aripdcem: I will not, so I remove the line 214.

aripddev:
Custom theme example is working fine when custom theme stylesheet is put resources directory of the webapp.
However, I am unable to access resources directory of primefaces-extensions to fetch ready-themes inside `keynote/themes`. I have lack of knowledge :)

aripddev:
There are two issues, I don't know how to solve them and need your help.

1. Nested `pe:keynoteItem` for vertical slides
2. Getting CSP error

EvalError: Refused to evaluate a string as JavaScript because 'unsafe-eval' is not an allowed source of script in the following Content Security Policy directive: "script-src 'self' 'nonce-MGI1ODZiOTgtNDgyZi00ZTI4LTlmZDYtYTNmMzhjMzhmZjc5'"

core/src/main/java/org/primefaces/extensions/component/keynote/Keynote.java
```diff
@@ -34,25 +34,40 @@
...
 @ResourceDependency(library = "primefaces-extensions", name = "keynote/keynote.js")
 @ResourceDependency(library = "primefaces-extensions", name = "keynote/keynote.css")
-public class Keynote extends AbstractDynamicData implements ClientBehaviorHolder, Widget {
+public class Keynote extends AbstractDynamicData implements Widget, ClientBehaviorHolder, PrimeClientBehaviorHolder {
 
     public static final String COMPONENT_TYPE = "org.primefaces.extensions.component.Keynote";
     public static final String COMPONENT_FAMILY = "org.primefaces.extensions.component";
     public static final String DEFAULT_RENDERER = "org.primefaces.extensions.component.KeynoteRenderer";
 
+    private static final String DEFAULT_EVENT = "end";
```
jepsar: Events should also be added to the taglib xml so they end up at the showcase documentation

core/src/main/resources/META-INF/resources/primefaces-extensions/keynote/widget.js
```diff
@@ -27,16 +27,48 @@ PrimeFaces.widget.ExtKeynote = PrimeFaces.widget.BaseWidget.extend({
...
-        deck.initialize();
+    /**
+     * Starts the keynote.
+     */
+    startKeynote: function () {
+        var $this = this;
 
-    }
+        $this.keynote.initialize().then(() => {
+            $this.keynote.on('slidetransitionend', event => {
```
jepsar: Why do you map this to "end" to only detect the last slide? Why not simply have a slidechanged event.

jepsar (changes requested):

core/src/main/java/org/primefaces/extensions/component/keynote/Keynote.java
aripdcem: I added it

core/src/main/resources/META-INF/resources/primefaces-extensions/keynote/widget.js
aripdcem: I removed "end", added `slidechanged` and `slidetransitionend`

core/src/main/java/org/primefaces/extensions/component/keynote/Keynote.java
jepsar: In the description I mean. See for example https://github.com/primefaces-extensions/primefaces-extensions/blob/68ff12efbfdf123a335d82e359d25b6222bf2b4e/core/src/main/resources/META-INF/primefaces-extensions.taglib.xml#L7908-L7909

core/src/main/java/org/primefaces/extensions/component/keynote/KeynoteRenderer.java
```diff
@@ -0,0 +1,296 @@
...
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public void encodeBegin(FacesContext context, UIComponent component) throws IOException {
+        final Keynote keynote = (Keynote) component;
+
+        // encodeCSS(context, "primefaces-extensions", "keynote/theme/" + keynote.getTheme() + ".css");
+        // encodeCSS(context, "primefaces", "primeicons/primeicons.css");
+        if (keynote.getTheme() != null && !"none".equals(keynote.getTheme())) {
```
jepsar:
```suggestion
        if (!"none".equals(keynote.getTheme())) {
```

core/src/main/java/org/primefaces/extensions/component/keynote/KeynoteRenderer.java
```diff
@@ -0,0 +1,296 @@
...
+
+        if (keynote.getVar() != null) {
+            // dynamic items
+            final Object value = keynote.getValue();
+            if (value != null) {
+                if (!(value instanceof Collection<?>)) {
+                    throw new FacesException("Value in Keynote must be of type Collection / List");
+                }
+
+                for (final UIComponent kid : keynote.getChildren()) {
```
jepsar: Use index loop https://github.com/primefaces/primefaces/wiki/Contributing-to-PrimeFaces#performance-guidelines

core/src/main/java/org/primefaces/extensions/component/keynote/KeynoteRenderer.java
```diff
@@ -0,0 +1,296 @@
...
+                    if (kid.isRendered() && !(kid instanceof UIKeynoteItem)) {
+                        // first render children like stamped elements, etc.
+                        renderChild(context, kid);
+                    }
+                }
+
+                final Collection<KeynoteItem> col = (Collection<KeynoteItem>) value;
+                for (final KeynoteItem keynoteItem : col) {
```
jepsar: Use index loop https://github.com/primefaces/primefaces/wiki/Contributing-to-PrimeFaces#performance-guidelines

core/src/main/java/org/primefaces/extensions/component/keynote/KeynoteRenderer.java
```diff
@@ -0,0 +1,296 @@
...
+        }
+
+        writer.startElement("div", null);
+        writer.writeAttribute(Attrs.CLASS, SLIDES_CLASS, "class");
+
+        if (keynote.getVar() != null) {
+            // dynamic items
+            final Object value = keynote.getValue();
+            if (value != null) {
+                if (!(value instanceof Collection<?>)) {
```
jepsar: This can simply be
```suggestion
                if (!(value instanceof Collection)) {
```
right?

core/src/main/java/org/primefaces/extensions/component/keynote/KeynoteRenderer.java
```diff
@@ -0,0 +1,296 @@
...
+
+                        // render item
+                        renderItem(context, writer, keynote, uiItem);
+                    }
+                }
+            }
+        }
+        else {
+            // static items
+            for (final UIComponent kid : keynote.getChildren()) {
```
jepsar: Use index loop https://github.com/primefaces/primefaces/wiki/Contributing-to-PrimeFaces#performance-guidelines

core/src/main/java/org/primefaces/extensions/component/keynote/KeynoteRenderer.java
```diff
@@ -0,0 +1,296 @@
...
+    }
+
+    /**
+     * Create the Javascript.
+     */
+    private void encodeScript(final FacesContext context, final Keynote keynote) throws IOException {
+        final WidgetBuilder wb = getWidgetBuilder(context);
+
+        wb.init("ExtKeynote", keynote)
+                    .attr("width", keynote.getWidth())
```
jepsar: Are there no default values for any of the attributes?

core/src/main/java/org/primefaces/extensions/component/keynote/KeynoteRenderer.java
```diff
@@ -0,0 +1,296 @@
...
+        encodeClientBehaviors(context, keynote);
+
+        wb.finish();
+    }
+
+    private void encodeCSS(FacesContext context, String library, String theme) throws IOException {
+        final ResponseWriter writer = context.getResponseWriter();
+        ExternalContext externalContext = context.getExternalContext();
+
+        Resource cssResource = context.getApplication().getResourceHandler().createResource(theme + ".css", library);
```
jepsar: @melloware I remember having issues with loading resources like this. I think we ended up loading them in JavaScript. Do you remember where?

core/src/main/java/org/primefaces/extensions/component/keynote/KeynoteRenderer.java
```diff
@@ -0,0 +1,296 @@
...
+        uiItem.encodeAll(context);
+
+        if (uiItem.isMarkdown()) {
+            writer.endElement("textarea");
+        }
+
+        if (keynote.isShowNotes()) {
+            writer.startElement("aside", null);
+            writer.writeAttribute(Attrs.CLASS, SPEAKER_NOTE_CLASS, null);
+            String note = uiItem.getNote() == null ? "" : uiItem.getNote();
```
jepsar:
```suggestion
            String note = uiItem.getNote() == null ? Constants.EMPTY_STRING : uiItem.getNote();
```

jepsar (changes requested):

jepsar:
Please update https://github.com/primefaces-extensions/primefaces-extensions/blob/master/core/src/main/resources/META-INF/NOTICE.txt

jepsar:
> There are two issues, I don't know how to solve them and need your help.
> 
> 1. Nested `pe:keynoteItem` for vertical slides

What is your issue here?

> 2. Getting CSP error
> 
> EvalError: Refused to evaluate a string as JavaScript because 'unsafe-eval' is not an allowed source of script in the following Content Security Policy directive: "script-src 'self' 'nonce-MGI1ODZiOTgtNDgyZi00ZTI4LTlmZDYtYTNmMzhjMzhmZjc5'"

Where are you getting this? Somewhere in your code?

core/src/main/resources/META-INF/resources/primefaces-extensions/keynote/widget.js
jepsar: I see you used their style, but I would prefer `slideChange` and `slideTransitionEnd`. See for example https://primefaces.github.io/primefaces/11_0_0/#/components/datatable?id=ajax-behavior-events

core/src/main/java/org/primefaces/extensions/component/keynote/KeynoteRenderer.java
melloware: Yes I feel like the issue was with inputphone and loading its language files?

melloware:
Somewhere in their code i wondering if they are doing `eval` a javascript string.  In PF I had to change these to a custom function I wrote so `PrimeFaces.csp.eval(xx);` which uses Jquery and the PF CSP NONCE to execute the eval.

core/src/main/java/org/primefaces/extensions/component/keynote/Keynote.java
aripdcem: Sorry, I misunderstood. It is corrected.

aripddev:
> > There are two issues, I don't know how to solve them and need your help.
> > 
> > 1. Nested `pe:keynoteItem` for vertical slides
> 
> What is your issue here?

For vertical slides, it is needed to generate sections like this;
```html
<section><!-- Horizontal slide tag -->
  <section>Vertical Slide 1</section><!-- Vertical slide tag -->
  <section>Vertical Slide 2</section><!-- Vertical slide tag -->
</section>
```
with `pe:keynoteItem` I can only generate horizontal tag. Here is the screenshot.

Screen Shot 2021-12-09 at 23 46 33

> 
> > 2. Getting CSP error
> > 
> > EvalError: Refused to evaluate a string as JavaScript because 'unsafe-eval' is not an allowed source of script in the following Content Security Policy directive: "script-src 'self' 'nonce-MGI1ODZiOTgtNDgyZi00ZTI4LTlmZDYtYTNmMzhjMzhmZjc5'"
> 
> Where are you getting this? Somewhere in your code?

Not in PFE showcase since CSP is not activated, but in my app that CSP is enabled.

core/src/main/java/org/primefaces/extensions/component/keynote/KeynoteRenderer.java
jepsar: In that case it must have been the utils script

core/src/main/java/org/primefaces/extensions/component/keynote/KeynoteRenderer.java
aripdcem: This is the similiar code in other PF/PFE components. Am I wrong?

jepsar: That may very well be, but that does not make it the best practice. Please read the linked section.

core/src/main/java/org/primefaces/extensions/component/keynote/KeynoteRenderer.java
jepsar: @aripddev can you please try what happens when you Ajax update a section of the page containing a keynote component toggling it from rendered false to true.

jepsar:
> with `pe:keynoteItem` I can only generate horizontal tag

You probably need to create a renderer for a `pe:keynoteItem` and in both `pe:keynote` and `pe:keynoteItem` render childeren.

core/src/main/java/org/primefaces/extensions/component/keynote/KeynoteRenderer.java
aripdcem: ok, done.

aripddev:
> Somewhere in their code i wondering if they are doing `eval` a javascript string. In PF I had to change these to a custom function I wrote so `PrimeFaces.csp.eval(xx);` which uses Jquery and the PF CSP NONCE to execute the eval.

@melloware There are two js files which are reveal.js and markdown.js (a plugin). There is no `eval` inside them.

melloware:
If you can get me the exact error and it should point to a line of code you can click on so it will help use figure out where exactly the eval call is. That would be great.

core/src/main/java/org/primefaces/extensions/component/keynote/KeynoteRenderer.java
aripdcem: Hmm. When I added Ajax update targeted to "btnSubmit" in my app, it throws below exception. It is working well with PFE showcase.

"Cannot find component for expression "btnSubmit" ....."

aripddev:
It is in reveal.js

Screen Shot 2021-12-10 at 14 58 45

melloware:
Yes that Function call is not allowed according to CSP.  That is similar to `eval` see the official docs here: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/script-src#unsafe_eval_expressions

> The 'unsafe-eval' source expression controls several script execution methods that create code from strings. If 'unsafe-eval' isn't specified with the script-src directive, the following methods are blocked and won't have any effect:
> 
> eval()
> Function()
> When passing a string literal like to methods like: window.setTimeout("alert(\"Hello World!\");", 500);
> setTimeout()
> setInterval()
> window.setImmediate
> window.execScript()  (IE < 11 only)

melloware:
I think you will have to report it to the Reveal.js team on their github.

melloware:
Looks like your real CSP issue is here: https://github.com/jaredpalmer/tsdx/issues/954

And core issue: https://github.com/facebook/regenerator/issues/378

aripddev:
> I think you will have to report it to the Reveal.js team on their github.

Reported [hakimel/reveal.js#3091](https://github.com/hakimel/reveal.js/issues/3091)

core/src/main/java/org/primefaces/extensions/component/keynote/KeynoteRenderer.java
```diff
@@ -0,0 +1,326 @@
...
+
+        if (keynote.getVar() != null) {
+            // dynamic items
+            final Object value = keynote.getValue();
+            if (value != null) {
+                if (!(value instanceof Collection)) {
+                    throw new FacesException("Value in Keynote must be of type Collection / List");
+                }
+
+                // TODO will be removed
```
jepsar: Will be delegated to `KeynoteItemRenderer`, right?

aripdcem: I tried something but failed. A little help would be nice. I need to create a new renderer class and register it to faces-config, right?

aripdcem: is it something like layout, layoutPane components?

aripdcem: It is done

core/src/main/java/org/primefaces/extensions/component/keynote/KeynoteRenderer.java
jepsar: I'll test it once merged

jepsar (approved):

jepsar:
Great job! Thanks!

jepsar:
@aripddev Is this ready for a final review?

aripddev:
@jepsar Except for CSP, there is one issue which I couldn't find the solution. The remain issue is to change theme, not custom but ready one.

aripddev:
@jepsar I am looking at the thematization of `codemirror` component. It seems that every themes have their own class names and all themes are loaded at once. In reveal.js, since each theme is available as a separate stylesheet, they all have the same class names. Loading all at once will not work. Am i right in my thinking?

I suggest to render the ready themes by changing the original style files and giving different class names for each, as in `codemirror` component. .reveal-white, .reveal-beige, ... so on. What do you think?

jepsar:
@aripddev in `InputPhone` I send a resource path to the widget:

https://github.com/primefaces-extensions/primefaces-extensions/blob/4cba61a86c7e5c974a3bbf89ec7fbe1da279ee92/core/src/main/java/org/primefaces/extensions/component/inputphone/InputPhoneRenderer.java#L210-L216

You could do that with CSS resource and inject it into the DOM in the widget.

aripddev:
@jepsar Removed `encodeCSS` method and added below lines to `encodeScript` method.

```java
160 if (!"none".equals(keynote.getTheme())) {
161     wb.attr("utilsScript",
162             context.getApplication()
163                     .getResourceHandler()
164                     .createResource("keynote/theme/" + keynote.getTheme() + ".css", "primefaces-extensions")
165                     .getRequestPath());
166 }
```

It throws

```sh
java.lang.NullPointerException
	at org.primefaces.extensions.component.keynote.KeynoteRenderer.encodeScript(KeynoteRenderer.java:165)
```

To see what happens I just changed it to `inputphone/utils.js`, it works fine.

```javascript
pf.cw("ExtKeynote", "widget_j_id_x", {
    id: "j_id_x",
    width: 480,
    height: 350,
    theme: "night",
    utilsScript: "\/showcase\-ext\/javax.faces.resource\/inputphone\/utils.js.jsf?ln=primefaces\-extensions\x26v=11.0.0\-SNAPSHOT\x26e=11.0.0\-SNAPSHOT"
});
```

jepsar:
@aripddev I'll have a closer look at it this evening. First guess is that the resources are not included in the JAR. Can you double check the POM please?

aripddev:
Thanks @jepsar . BTW all themes provided by reveal.js are inside keynote directory.

Screen Shot 2021-12-13 at 17 11 01

jepsar:
@aripddev are you on Discord? Could you add me? Jasper#0844

jepsar:
See https://github.com/jepsar/primefaces-extensions/commit/37f967553c7e04e460547cca6f9a769964c8c729

core/src/main/java/org/primefaces/extensions/component/keynote/Keynote.java
```diff
@@ -0,0 +1,613 @@
...
+        navigationMode,
+        progress,
+        showNotes,
+        slideNumber,
+        touch,
+        transition,
+        transitionSpeed,
+        backgroundTransition,
+        theme,
+        library,
```
jepsar, line 97: @melloware what do you think of this way of offering themes to be loaded?

IMO we should just offer / support the bundled themes. And if you want something custom, simply create custom styles in a custom style sheet.

jepsar:
@melloware @aripddev What is the status here? Are we waiting for hakimel/reveal.js#3091?

melloware:
I don't think we should wait for that CSP fix to merge this.   If everything else is in good shape @aripddev ?

aripddev:
@jepsar , @melloware It is ok for me. Once CSP issue fixes we can update it. Thanks.

melloware (approved):

PR #660, opened 2021-12-18
MorphiaLazyDataModel: Implements some of the missing match modes

### LINKED ISSUE #659: New Helper: Mongo DB Lazy Datatable implementation

newk5:
Implements the following missing match modes:
- BETWEEN
- NOT_CONTAINS
- NOT_EQUALS
- NOT_STARTS_WITH
- NOT_ENDS_WITH
- NOT_IN
- GLOBAL (this one is expected to implemented by the user using the new .globalFilter(BiConsumer) method, since this wouldn't be possible to implement in a generic manner without having deep knowledge of the model being used)

I think this covers all the cases for this datamodel to be used in any scenario when morphia is used, making it unnecessary to ever need to create a custom morphia DataModel.

For cases when a complex field filter is needed which is not covered by the match modes on this datamodel, use:
- overrideFieldQuery(field, BiConsumer<Query, FilterMeta>)

For cases where you need to add more stuff to the query that doesn't depend on filters use:
- prependQuery(Consumer<Query>)

For cases where you want to implment a globalFiter use:
- globalFilter(BiConsumer<Query, FilterMeta>)

### DIFF
--- a/core/src/main/java/org/primefaces/extensions/model/mongo/MorphiaLazyDataModel.java
+++ b/core/src/main/java/org/primefaces/extensions/model/mongo/MorphiaLazyDataModel.java
@@ -262,7 +262,7 @@ public class MorphiaLazyDataModel<T> extends LazyDataModel<T> implements Seriali
                                 }
                                 break;
                             case NOT_CONTAINS:
-                                 q.filter(Filters.regex(field).pattern(val + "").caseInsensitive().not());
+                                 q.filter(Filters.regex(field).pattern(val + Constants.EMPTY_STRING).caseInsensitive().not());
 
                                 break;
                                 
--- a/core/src/main/java/org/primefaces/extensions/model/mongo/MorphiaLazyDataModel.java
+++ b/core/src/main/java/org/primefaces/extensions/model/mongo/MorphiaLazyDataModel.java
@@ -254,7 +254,7 @@ public class MorphiaLazyDataModel<T> extends LazyDataModel<T> implements Seriali
                                 
                             case BETWEEN:
                                 if (metadata.getFilterValue().getClass() == ArrayList.class) {
-                                    final List dates = (ArrayList) metadata.getFilterValue();
+                                    final List<?> dates = (List) metadata.getFilterValue();
                                     if (dates.size() > 1){ //does this ever have less than 2 items?
                                         q.filter(Filters.gt(field, dates.get(0)),Filters.lt(field,dates.get(1)));
                                     }
--- a/core/src/main/java/org/primefaces/extensions/model/mongo/MorphiaLazyDataModel.java
+++ b/core/src/main/java/org/primefaces/extensions/model/mongo/MorphiaLazyDataModel.java
@@ -285,7 +285,7 @@ public class MorphiaLazyDataModel<T> extends LazyDataModel<T> implements Seriali
                                 break;
                             case NOT_IN:
                                 
-                                if (metadata.getFilterValue().getClass() == Object[].class) {
+                                if (metadata.getFilterValue() instanceof Object[]) {
                                     final Object[] parts = (Object[]) metadata.getFilterValue();
                                     q.filter(Filters.nin(field, Arrays.asList(parts)));
                                 }
--- a/core/src/main/java/org/primefaces/extensions/model/mongo/MorphiaLazyDataModel.java
+++ b/core/src/main/java/org/primefaces/extensions/model/mongo/MorphiaLazyDataModel.java
@@ -253,7 +253,7 @@ public class MorphiaLazyDataModel<T> extends LazyDataModel<T> implements Seriali
                                 break;
                                 
                             case BETWEEN:
-                                if (metadata.getFilterValue().getClass() == ArrayList.class) {
+                                if (metadata.getFilterValue() instanceof List) {
                                     final List<?> dates = (List) metadata.getFilterValue();
                                     if (dates.size() > 1){ //does this ever have less than 2 items?
                                         q.filter(Filters.gt(field, dates.get(0)),Filters.lt(field,dates.get(1)));
--- a/core/src/main/java/org/primefaces/extensions/model/mongo/MorphiaLazyDataModel.java
+++ b/core/src/main/java/org/primefaces/extensions/model/mongo/MorphiaLazyDataModel.java
@@ -40,6 +40,7 @@ import javax.faces.FacesException;
 import org.primefaces.model.FilterMeta;
 import org.primefaces.model.LazyDataModel;
 import org.primefaces.model.SortMeta;
+import org.primefaces.util.Constants;
 import org.primefaces.util.Lazy;
 
 import dev.morphia.Datastore;
@@ -48,7 +49,6 @@ import dev.morphia.query.Query;
 import dev.morphia.query.Sort;
 import dev.morphia.query.experimental.filters.Filters;
 import dev.morphia.query.experimental.filters.RegexFilter;
-import java.util.ArrayList;
 
 /**
  * Basic {@link LazyDataModel} implementation for MongoDB using Morphia.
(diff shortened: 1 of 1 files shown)


melloware:
Don't forget to run `mvn spotless:apply` before checking in the build is failing

core/src/main/java/org/primefaces/extensions/model/mongo/MorphiaLazyDataModel.java
```diff
@@ -248,6 +251,61 @@ public int count(final Map<String, FilterMeta> map) {
                                 }
 
                                 break;
+                                
+                            case BETWEEN:
+                                if (metadata.getFilterValue().getClass() == ArrayList.class) {
+                                    final List dates = (ArrayList) metadata.getFilterValue();
+                                    if (dates.size() > 1){ //does this ever have less than 2 items?
+                                        q.filter(Filters.gt(field, dates.get(0)),Filters.lt(field,dates.get(1)));
```
jepsar: Shouldn't this be `gte`?
```suggestion
                                        q.filter(Filters.gte(field, dates.get(0)), Filters.lt(field,dates.get(1)));
```

core/src/main/java/org/primefaces/extensions/model/mongo/MorphiaLazyDataModel.java
```diff
@@ -248,6 +251,61 @@ public int count(final Map<String, FilterMeta> map) {
                                 }
 
                                 break;
+                                
+                            case BETWEEN:
+                                if (metadata.getFilterValue().getClass() == ArrayList.class) {
+                                    final List dates = (ArrayList) metadata.getFilterValue();
```
jepsar:
```suggestion
                                    final List<?> dates = (List) metadata.getFilterValue();
```

core/src/main/java/org/primefaces/extensions/model/mongo/MorphiaLazyDataModel.java
```diff
@@ -248,6 +251,61 @@ public int count(final Map<String, FilterMeta> map) {
...
+                                    if (dates.size() > 1){ //does this ever have less than 2 items?
+                                        q.filter(Filters.gt(field, dates.get(0)),Filters.lt(field,dates.get(1)));
+                                    }
+                                  
+                                }
+                                break;
+                            case NOT_CONTAINS:
+                                 q.filter(Filters.regex(field).pattern(val + "").caseInsensitive().not());
```
jepsar:
```suggestion
                                 q.filter(Filters.regex(field).pattern(val + Constants.EMPTY_STRING).caseInsensitive().not());
```

core/src/main/java/org/primefaces/extensions/model/mongo/MorphiaLazyDataModel.java
```diff
@@ -248,6 +251,61 @@ public int count(final Map<String, FilterMeta> map) {
...
+                            case NOT_STARTS_WITH:
+                                
+                                final RegexFilter regStartsWithNot = Filters.regex(field);
+                                regStartsWithNot.pattern("^" + val).caseInsensitive();
+                                q.filter(regStartsWithNot.not());
+
+                                break;
+                            case NOT_IN:
+                                
+                                if (metadata.getFilterValue().getClass() == Object[].class) {
```
jepsar:
```suggestion
                                if (metadata.getFilterValue() instanceof Object[]) {
```

core/src/main/java/org/primefaces/extensions/model/mongo/MorphiaLazyDataModel.java
```diff
@@ -248,6 +251,61 @@ public int count(final Map<String, FilterMeta> map) {
                                 }
 
                                 break;
+                                
+                            case BETWEEN:
+                                if (metadata.getFilterValue().getClass() == ArrayList.class) {
```
jepsar:
```suggestion
                                if (metadata.getFilterValue() instanceof List) {
```

core/src/main/java/org/primefaces/extensions/model/mongo/MorphiaLazyDataModel.java
newk5: At first I was using gte and lte to include both the start and end date, but I wasn't sure if we should  make the assumption that between queries include the start and end date, so I chose to just use lt and gt and not include them.

I dont have a strong preference one way or another, should between be date inclusive?

jepsar: Maybe apples and pears, but in ANSI SQL they are both inclusive. But, with dates you might want to exclude the end date as explained here https://stackoverflow.com/questions/749615/does-ms-sql-servers-between-include-the-range-boundaries/749663

newk5: Hmm yea in mongo there is no between keyword, "between queries" are usually done with lt, lte, gt, gte. That said, I was now looking at the "between" match mode for non-lazy datatables on the showcase here: https://www.primefaces.org/showcase/ui/data/datatable/filter.xhtml?jfwid=abd89

And I noticed that it includes the start and end date, so maybe we should match that and make this query include the dates aswell? What do you think?

jepsar: Yes, I would match the showcase, so both inclusive. @melloware WDYT?

Maybe we should break up each mode into separate methods, so they are easily overridable. So, you can extend this class and only override the between filter method (or whatever method you want to change).

melloware: Yes we should match the showcase and PF original intent so if that is what it's doing it's what we should do too.

newk5: That sounds good too.

I had an idea the other day which I was going to propose, although it's not PFE related, it depends on PF, but since we're on the subject... 

It would be great if we allowed the user to specify a custom match mode. At the moment if you specify a match mode that doesn't exist, the filterMeta matchmode returns null. It would be nice if it returned something like "CUSTOM" and if we had a way to then access the matchmode name specified on the p:column's filterMatchMode="". That way, the generic data models like this one and JPA's could have a method like .matchMode("customMatchModeName", BiConsumer< Query, FilterMeta > which would allow the user to implement their custom match mode.

That said, I'm not familiar with PF's internal matchmode implementation and dont know if the way the current structure is set up would allow for such a change.

melloware: PrimeReact has a CUSTOM so I don't see why we cant add it to PF

newk5: Alright I changed it so between is date inclusive and applied the spotless formatter, I think the build is no longer failing. I dont think I'll have much free time for the next few days so I'm not sure when I'll get to split up each match mode into separate methods, but maybe we can leave that for a future PR if you all agree

core/src/main/java/org/primefaces/extensions/model/mongo/MorphiaLazyDataModel.java
```diff
@@ -249,6 +252,59 @@ public int count(final Map<String, FilterMeta> map) {
...
+                                    if (dates.size() > 1) { // does this ever have less than 2 items?
+                                        q.filter(Filters.gte(field, dates.get(0)), Filters.lte(field, dates.get(1)));
+                                    }
+
+                                }
+                                break;
+                            case NOT_CONTAINS:
+                                q.filter(Filters.regex(field).pattern(val + Constants.EMPTY_STRING).caseInsensitive().not());
+
+                                break;
```
jepsar, line 267: Technically this passed Spotless, but can you please align blank lines with the other filters?

melloware (approved):

melloware:
I am going to merge this and then make Thomas's recommended changes

PR #766, opened 2022-03-23
Remove unused dependency url-loader #765

### LINKED ISSUE #765: Monaco: Replace url-loader with asset modules
URL Loader is deprecated and no longer supported: https://github.com/webpack-contrib/url-loader

They recommend switching to Assert Modules: https://webpack.js.org/guides/asset-modules/

cc @blutorange

blutorange:
Seems like I had already switched to the assets-module in the webpack.config, so the url-loader dependency wasn't even being used anymore. Sometimes life can be so easy!


melloware:
Awesome thanks!

melloware (approved):

PR #776, opened 2022-03-26
Fix #771: CookiePolicy showcase

### LINKED ISSUE #771: New Component: CookiePolicy
Within our project we created a cookiePolicy component to be able to handle acceptance of cookie policies.
Example usage:
```xml
<pe:cookiePolicy>
    Do you agree with our cookie policy?
    <p:commandButton value="Only essential cookies">
        <pe:saveCookiePolicy policy="essential" retention="#{60*60*24*30}"/>
    </p:commandButton>
    <p:commandButton value="Accept all">
        <pe:saveCookiePolicy policy="essential, statistics" retention="#{60*60*24*30}"/>
    </p:commandButton>
</pe:cookiePolicy>
```
Here `pe:cookiePolicy` will only render its children when no cookie policy has been accepted before. Accepted cookie policies are stored within a cookie itself.
Via the `pe:saveCookiePolicy` action one can save the preferred cookie policy.

If there is interest in such component as part of `primefaces-extensions`, please make some noise here.

fcorneli:


jepsar (approved):

jepsar:
@fcorneli great work! Thanks!

@melloware I'll merge now and will have a look (at the image zoom as well)

PR #788, opened 2022-04-03
Fixes #785, apply value upon initialization

### LINKED ISSUE #785: MonacoEditor: Value set programatically before render not applied correctly during rendering
Sounds a bit complicated, you can assign this to me. Fix is rather simple.

---

Monaco editor is initialized asynchronously. The monaco editor widget has a method `setValue` / `setValueNow`. When the editor was not initialized yet, those methods store the target value and set that value once the editor was initialized.

The value is also set on the model when the model is created, but when the monaco editor is not visible on page load  (e.g. when it is inside a tab), the editor is rendered only once it becomes visible. If `setValue` / `setValueNow` is called in that interval, the new value is not set correctly once the editor is rendered. 

https://github.com/primefaces-extensions/primefaces-extensions/blob/master/core/src/main/resources/META-INF/resources/primefaces-extensions/monacoeditor/03-monaco-widget-base.js#L225-L227

The `setValue` call on the lines above does nothing since the `initialized` data key is only set afterward. We need to change the order of these lines, or call another method to ensure the value is being updated.

---

Can be worked around either by replacing `widget.setValueNow(newValue)` with `widget.whenReady().then(e => e.setValueNow(newValue))`, by using the `initialized` behavior event to set the value, or by using one of the extender methods to set the value directly on the model.

blutorange:
Simple fixes are the best. By setting the `initialized` attribute before calling `setValue`, said function actually sets the value instead of just remembering it for later.

Can be tested e.g. in the playground (https://www.primefaces.org/showcase-ext/sections/monacoEditor/extender.jsf?example=jquery) by calling `setValueNow` in the `afterCreate` callback of the extender. Before this fix, the value was not applied, now it is

```javascript
/**
 * @return {PrimeFaces.widget.ExtMonacoEditor.ExtenderCodeEditorInline}
 */
function createExtender() {
  return {
    async afterCreate(context) {
      context.setValueNow("const helloWord = 42");
    }
  };
}
```

### DIFF
--- a/core/src/main/resources/META-INF/resources/primefaces-extensions/monacoeditor/03-monaco-widget-base.js
+++ b/core/src/main/resources/META-INF/resources/primefaces-extensions/monacoeditor/03-monaco-widget-base.js
@@ -222,9 +222,9 @@ window.monacoModule = window.monacoModule || {};
      * @returns {Promise<void>}
      */
     async _onInitSuccess() {
+      this.jq.data("initialized", true);
       await this.setValue(this._editorValue || "");
       this._fireEvent("initialized");
-      this.jq.data("initialized", true);
       for (const { resolve } of this._onDone || []) {
         resolve(this);
       }


melloware (approved):

melloware:
Nice simple fix!

PR #887, opened 2022-07-19
Npe with null values

### LINKED ISSUE #886: Sheet: npe when putting null value
### Describe the bug

PR #852 introduced an NPE on `Sheet.setLocalValue(Sheet.java:294)`, ConcurrentHashMap does not allow null for value:
```
    final V putVal(K key, V value, boolean onlyIfAbsent) {
        if (key == null || value == null) throw new NullPointerException();
```

### Reproducer

_No response_

### Expected behavior

Trying to put a null value into localValues should not result in an npe. Instead:

1. nothing should be put, or
2. when the key is present the value should be removed

Before PR #852 null values were allowed in HasMap, so the second option seems the better one.

11.0.5-SNAPSHOT

### JSF implementation

Mojarra

### JSF version

2.3

### Browser(s)

all

eduarddrenth:
fixes #886


melloware:
Some checkstyle issues...

```
src/main/java/org/primefaces/extensions/component/sheet/Sheet.java:[295,9] (blocks) NeedBraces: 'if' construct must use '{}'s.
Error:  src/main/java/org/primefaces/extensions/component/sheet/Sheet.java:[295,18] (whitespace) WhitespaceAround: '!=' is not followed by whitespace.
Error:  src/main/java/org/primefaces/extensions/component/sheet/Sheet.java:[295,18] (whitespace) WhitespaceAround: '!=' is not preceded with whitespace.
Error:  src/main/java/org/primefaces/extensions/component/sheet/Sheet.java:[[297](https://github.com/primefaces-extensions/primefaces-extensions/runs/7410578088?check_suite_focus=true#step:8:298),9] (blocks) NeedBraces: 'else' construct must use '{}'s.
```

melloware:
one more `src/main/java/org/primefaces/extensions/component/sheet/Sheet.java:[297,9] (blocks) RightCurly: '}' at column 9 should be alone on a line.`

melloware (approved):

melloware:
I will just fix it

PR #888, opened 2022-07-19
11.0.x

### LINKED ISSUE #886: Sheet: npe when putting null value

eduarddrenth:
fixes #886 in 11.0.x branche


melloware:
Its OK I will cherry pick this commit from master to the 11.X branch no worries

PR #1031, opened 2023-01-14
Fixes #1030 Update rimraf

blutorange:
The only relevant change was that rimraf now supports promises directly, so I just had to remove the promise wrapper from the code

Also, happy new year by the way!


melloware:
Thank you as always!

PR #1129, opened 2023-04-01
#1112 TimeAgo future support

### LINKED ISSUE #1112: TimeAgo: Add future support
I was told you should be able to configure this in the jQuery plugin we are using.

jepsar:
Fix #1112

### DIFF
--- a/core/src/main/resources/META-INF/resources/primefaces-extensions/timeago/0-jquery.timeago.js
+++ b/core/src/main/resources/META-INF/resources/primefaces-extensions/timeago/0-jquery.timeago.js
@@ -42,7 +42,7 @@
     settings: {
       refreshMillis: 60000,
       allowPast: true,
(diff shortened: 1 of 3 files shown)


core/src/main/resources/META-INF/resources/primefaces-extensions/timeago/0-jquery.timeago.js
```diff
@@ -42,7 +42,7 @@
     settings: {
       refreshMillis: 60000,
       allowPast: true,
-      allowFuture: false,
+      allowFuture: true, // PFE #1112
```
jepsar, line 45: @melloware this was the easiest fix. IMO it makes sense to set this as the global default. If you don't this would display "The next month is 29 days ago", which is a bug in my book.

melloware (approved):
Agreed

PR #1133, opened 2023-04-03
SunEditor

### LINKED ISSUE #1110: New component: SunEditor
Pure javscript based WYSIWYG web editor, with no dependencies.

MIT Licensed: https://github.com/JiHong88/suneditor

To replace the removed CKEditor for WYSIWYG editing.

- [ ] Theme switching https://github.com/JiHong88/suneditor/issues/978
- [x] Sanitizing like TextEditor
- [x] Extender support
- [x] Test inside dialog
- [x] Right-To-Left Support
- [x] Button assignment using JSON?
- [x] Showcase examples including Locale switching

melloware:
Fix #1110

@jepsar @newk5 and others here is initial version of the SunEditor courtesy of MattSho.

I added some features like RTL, Extender, and a few others.  But this needs review and Showcase examples!

- [x] Showcase basic example
- [x] Showcase `localizatron` example
- [x] Showcase `rtl` example
- [x] Showcase `mode` example
- [x] Showcase `buttons` example
- [x] Review ButtonListJSON. Not thrilled with that property or name (renamed toolbar)
- [x] Test in Dialog
- [x] Client Behavior Events Documented
- [ ] Theme switching https://github.com/JiHong88/suneditor/issues/978


core/src/main/java/org/primefaces/extensions/component/suneditor/SunEditor.java
```diff
@@ -0,0 +1,208 @@
...
+ */
+
+@ResourceDependency(library = "primefaces", name = "jquery/jquery.js")
+@ResourceDependency(library = "primefaces", name = "jquery/jquery-plugins.js")
+@ResourceDependency(library = "primefaces", name = "core.js")
+@ResourceDependency(library = "primefaces-extensions", name = "primefaces-extensions.js")
+@ResourceDependency(library = "primefaces-extensions", name = "suneditor/suneditor.min.css")
+@ResourceDependency(library = "primefaces-extensions", name = "suneditor/suneditor.min.js")
+@ResourceDependency(library = "primefaces-extensions", name = "suneditor/suneditor.css")
+@ResourceDependency(library = "primefaces-extensions", name = "suneditor/suneditor.js")
```
jepsar, line 49: Do we want to combine the JS into one file?

core/src/main/java/org/primefaces/extensions/component/suneditor/SunEditor.java
```diff
@@ -0,0 +1,208 @@
...
+    public String getWidgetVar() {
+        return (String) getStateHelper().eval(PropertyKeys.widgetVar, null);
+    }
+
+    public void setWidgetVar(String widgetVar) {
+        getStateHelper().put(PropertyKeys.widgetVar, widgetVar);
+    }
+
+    public int getWidth() {
+        return (Integer) getStateHelper().eval(PropertyKeys.width, Integer.MIN_VALUE);
```
jepsar, line 103: Do we want a width? In `p:textEditor` we just have height.

core/src/main/resources/META-INF/primefaces-extensions.taglib.xml
```diff
@@ -10135,4 +10135,222 @@ See showcase for an example. Default null.]]>
...
+                <![CDATA[Name of javascript function to extend the options of the underlying SunEditor configuration.]]>
+            </description>
+            <name>extender</name>
+            <required>false</required>
+            <type>java.lang.String</type>
+        </attribute>
+        <attribute>
+            <description>
+                <![CDATA[json array of array using single quote that defines the toolbar]]></description>
+            <name>ButtonListJSON</name>
```
jepsar: Name does not meet conventions. Why not simply call it `toolbar`?

core/src/main/resources/META-INF/resources/primefaces-extensions/suneditor/suneditor-widget.js
```diff
@@ -0,0 +1,172 @@
...
+        if (this.disabled) {
+            this.editor.disable();
+        }
+
+        //update input on change
+        this.editor.onChange = function(contents, core) {
+            $this.input.val([].map.call(core.context.element.wysiwyg.children, function(node) {
+                return node.outerHTML || "";
+            }).join(""));
+            $this.callBehavior('change');
```
jepsar: The behaviors are not documented in the taglib xml.

core/src/main/resources/META-INF/resources/primefaces-extensions/suneditor/suneditor-widget.js
```diff
@@ -0,0 +1,172 @@
...
+     * @param {object} cfg The widget configuration.
+     */
+    init: function(cfg) {
+        this._super(cfg);
+        this.input = $(this.jqId);
+        this.disabled = (cfg.disabled === undefined) ? false : cfg.disabled;
+
+        if (this.disabled) {
+            this.input.attr("disabled", "disabled");
+            this.jq.addClass("ui-state-disabled");
```
jepsar: Shouldn't this go to the enable / disable functions?

jepsar:
Oh, I've overlooked one major thing in this PR. Can you please add it to the showcase?

core/src/main/resources/META-INF/resources/primefaces-extensions/suneditor/suneditor-widget.js
melloware: I don;t think these are `onChange` client behaviors they are AJAX `event="change"` behaviors so they don't go in the taglib right?

core/src/main/resources/META-INF/primefaces-extensions.taglib.xml
melloware: Agreed I like toolbar better

core/src/main/java/org/primefaces/extensions/component/suneditor/SunEditor.java
melloware: Yes this was my comment about not using a separate module `resources-suneditor` I can then combine these as a normal widget.

core/src/main/java/org/primefaces/extensions/component/suneditor/SunEditor.java
melloware: Wasn't sure the original dev MattSho did this I assume so you could control the exact size of the editor.

melloware:
Yeah its on the TODO.  was hoping to get help with Showcase, Dark mode, etc.

See the checklist above. 😄

core/src/main/resources/META-INF/resources/primefaces-extensions/suneditor/suneditor-widget.js
jepsar: Ajax events should end up in the documentation right? We get them from the taglib if I'm right.

Screenshot 2023-04-04 at 15 05 42

```xml
        <description>
            <![CDATA[InputPhone is an extension to the inputText with optimized handling for phone numbers.
     Client behavior events: countrySelect - fires when a country is selected (org.primefaces.event.SelectEvent).]]>
        </description>
```

jepsar:
> Yeah its on the TODO.

Okay.. don't know how I missed that 🙈

> was hoping to get help with Showcase, Dark mode, etc.

I can take care of it

core/src/main/resources/META-INF/resources/primefaces-extensions/suneditor/suneditor-widget.js
melloware: Oh I forgot we had magic to do this...

melloware:
OK I combined the JS so now its a normal  widget, renamed the JSON property to just `toolbar`.

core/src/main/resources/META-INF/resources/primefaces-extensions/suneditor/suneditor-widget.js
melloware: OK so TextEditor in PF does this too.  Its so it never bothers Creating the QUillJS object if the component was disabled so it renders the DIV and TextArea Disabled without even calling SUNEDITOR to create the editor.  He followed what PF TextEditor is doing for right or wrong?

jepsar: Alright. So it seems there is a small bug in PF then.

core/src/main/resources/META-INF/resources/primefaces-extensions/suneditor/suneditor-widget.js
jepsar: I'll add the events to the showcase tomorrow as well as a locale.

melloware: OK cool the locales I include in the JS under /lang we just need to figure out how to load one!  I think it might be similar to what you have done in other plugins.

melloware:
Also I wonder if we can just combine and include all the language files as well. And the locale will just pick the right language.

melloware:
OK added showcase examples for Localization, RTL, and Mode.

I think we just need one for custom button list.

jepsar:
I think we are good to merge. Only thing is that the basic example is not really basic. 😅

jepsar (approved):

melloware:
Love it!

PR #1149, opened 2023-04-10
Update mkdirp and rimraf

blutorange:
This contains the commits from both #1147 and #1148 and fixes the imports from the mkdirp and rimraf update. I also tested it with the showcase, Monaco Editor is still working.


melloware:
Awesome!  Thanks for fixing!

PR #1424, opened 2024-01-24
#283 New Component: ClockPicker

### LINKED ISSUE #283: New Component: TimeClockPicker
Screen Shot 2021-01-24 at 16 00 39

jxmai:
Fixes https://github.com/primefaces-extensions/primefaces-extensions/issues/283

Uses: https://weareoutman.github.io/clockpicker/

Sample usage screenshot:

image

image

### DIFF
--- a/core/src/main/java/org/primefaces/extensions/component/clockpicker/ClockPicker.java
+++ b/core/src/main/java/org/primefaces/extensions/component/clockpicker/ClockPicker.java
@@ -35,9 +35,8 @@ import org.primefaces.util.LocaleUtils;
 @ResourceDependency(library = "primefaces", name = "jquery/jquery.js")
 @ResourceDependency(library = "primefaces", name = "jquery/jquery-plugins.js")
 @ResourceDependency(library = "primefaces", name = "core.js")
-@ResourceDependency(library = "primefaces-extensions", name = "clockpicker/0-clockpicker.css")
-@ResourceDependency(library = "primefaces-extensions", name = "clockpicker/0-clockpicker.js")
-@ResourceDependency(library = "primefaces-extensions", name = "clockpicker/1-clockpicker-widget.js")
+@ResourceDependency(library = "primefaces-extensions", name = "clockpicker/clockpicker.css")
+@ResourceDependency(library = "primefaces-extensions", name = "clockpicker/clockpicker.js")
 public class ClockPicker extends HtmlInputText implements Widget {
     public static final String CONTAINER_CLASS = "pe-clockpicker ui-widget ui-corner-all input-group clockpicker";
 
--- a/core/src/main/java/org/primefaces/extensions/component/clockpicker/ClockPicker.java
+++ b/core/src/main/java/org/primefaces/extensions/component/clockpicker/ClockPicker.java
@@ -1,0 +1,154 @@
+/*
+ * Copyright (c) 2011-2024 PrimeFaces Extensions
+ *
+ *  Permission is hereby granted, free of charge, to any person obtaining a copy
+ *  of this software and associated documentation files (the "Software"), to deal
+ *  in the Software without restriction, including without limitation the rights
+ *  to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ *  copies of the Software, and to permit persons to whom the Software is
+ *  furnished to do so, subject to the following conditions:
+ *
+ *  The above copyright notice and this permission notice shall be included in
+ *  all copies or substantial portions of the Software.
+ *
+ *  THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ *  IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ *  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ *  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ *  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ *  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ *  THE SOFTWARE.
+ */
+package org.primefaces.extensions.component.clockpicker;
+
+import java.util.Locale;
+
+import javax.faces.application.ResourceDependency;
+import javax.faces.component.html.HtmlInputText;
+import javax.faces.context.FacesContext;
+
+import org.primefaces.component.api.Widget;
+import org.primefaces.util.Constants;
+import org.primefaces.util.LocaleUtils;
+
+@ResourceDependency(library = "primefaces", name = "components.css")
+@ResourceDependency(library = "primefaces", name = "jquery/jquery.js")
+@ResourceDependency(library = "primefaces", name = "jquery/jquery-plugins.js")
+@ResourceDependency(library = "primefaces", name = "core.js")
+@ResourceDependency(library = "primefaces-extensions", name = "clockpicker/0-clockpicker.css")
+@ResourceDependency(library = "primefaces-extensions", name = "clockpicker/0-clockpicker.js")
+@ResourceDependency(library = "primefaces-extensions", name = "clockpicker/1-clockpicker-widget.js")
+public class ClockPicker extends HtmlInputText implements Widget {
+    public static final String CONTAINER_CLASS = "pe-clockpicker ui-widget ui-corner-all input-group clockpicker";
+
+    public static final String COMPONENT_TYPE = "org.primefaces.extensions.component.ClockPicker";
+    public static final String COMPONENT_FAMILY = "org.primefaces.extensions.component";
+    private static final String DEFAULT_RENDERER = "org.primefaces.extensions.component.ClockPickerRenderer";
+
+    private Locale appropriateLocale;
+
+    protected enum PropertyKeys {
+
+        //@formatter:off
+        widgetVar,
+        placement,
+        align,
+        donetext,
+        autoclose,
+        locale,
+        vibrate;
+
+        private String toString;
+
+        PropertyKeys(final String toString) {
+            this.toString = toString;
+        }
+
+        PropertyKeys() {
+        }
+
+        @Override
+        public String toString() {
+            return toString != null ? toString : super.toString();
+        }
+    }
+    
+    public ClockPicker() {
+        setRendererType(DEFAULT_RENDERER);
+    }
+    
+    
+    private boolean isSelfRequest(final FacesContext fc) {
+        return this.getClientId(fc).equals(
+                    fc.getExternalContext().getRequestParameterMap().get(Constants.RequestParams.PARTIAL_SOURCE_PARAM));
+    }
+
+    @Override
+    public String getFamily() {
+        return COMPONENT_FAMILY;
+    }
+    
+    public String getWidgetVar() {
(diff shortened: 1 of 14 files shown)


core/src/main/java/org/primefaces/extensions/component/clockpicker/ClockPickerRenderer.java
```diff
@@ -0,0 +1,200 @@
...
+            }
+        }
+    }
+
+    @Override
+    public Object getConvertedValue(FacesContext context, UIComponent component,
+                Object value) throws ConverterException {
+        final ClockPicker clockPicker = (ClockPicker) component;
+        String submittedValue = (String) value;
+        SimpleDateFormat format = null;
```
jepsar: I would have used `LocalTime` as the default type. It's better suited for time only and easier to work with.

core/src/main/java/org/primefaces/extensions/component/clockpicker/ClockPicker.java
```diff
@@ -0,0 +1,154 @@
...
+import org.primefaces.util.Constants;
+import org.primefaces.util.LocaleUtils;
+
+@ResourceDependency(library = "primefaces", name = "components.css")
+@ResourceDependency(library = "primefaces", name = "jquery/jquery.js")
+@ResourceDependency(library = "primefaces", name = "jquery/jquery-plugins.js")
+@ResourceDependency(library = "primefaces", name = "core.js")
+@ResourceDependency(library = "primefaces-extensions", name = "clockpicker/0-clockpicker.css")
+@ResourceDependency(library = "primefaces-extensions", name = "clockpicker/0-clockpicker.js")
+@ResourceDependency(library = "primefaces-extensions", name = "clockpicker/1-clockpicker-widget.js")
```
jepsar: Please aggregate the scripts in the pom.xml. See for example https://github.com/primefaces-extensions/primefaces-extensions/blob/ad5024fb95812bb51a55ee82fa9b1c6a2c615d50/core/pom.xml#L194-L206

jepsar (changes requested):
Great work! Thanks! I don't have time to do a full review now, but I found two issues after a quick inspection.

core/src/main/java/org/primefaces/extensions/component/clockpicker/ClockPicker.java
melloware: Agreed if these are not minified you can just combine themwith `clockpicker/**`

core/src/main/java/org/primefaces/extensions/component/clockpicker/ClockPickerRenderer.java
```diff
@@ -0,0 +1,200 @@
...
+        writer.startElement("div", clockPicker);
+        writer.writeAttribute("class", ClockPicker.CONTAINER_CLASS, null);
+
+        writer.startElement("input", null);
+        writer.writeAttribute("id", inputId, null);
+        writer.writeAttribute("name", inputId, null);
+        writer.writeAttribute("type", "text", null);
+        writer.writeAttribute("class", "ui-inputfield ui-widget ui-state-default ui-corner-all", null);
+        writer.writeAttribute("size", 5, null);
+        writer.writeAttribute("maxlength", 5, null);
```
melloware, line 81: I am assuming size and maxlength must be 5 because that is the most there can be?  What about AM/PM?

core/src/main/java/org/primefaces/extensions/component/clockpicker/ClockPickerRenderer.java
```diff
@@ -0,0 +1,200 @@
...
+            return null;
+        }
+        else {
+            if (clockPicker.getConverter() != null) {
+                // convert via registered converter
+                return clockPicker.getConverter().getAsString(context, clockPicker, value);
+            }
+            else {
+                // use built-in converter
+                SimpleDateFormat timeFormat;
```
melloware: Combine these two lines `SimpleDateFormat timeFormat = = new SimpleDateFormat("HH:mm", clockPicker.calculateLocale());`

jepsar: You cannot use that with `LocalTime` anyway, right?

core/src/main/java/org/primefaces/extensions/component/clockpicker/ClockPickerRenderer.java
```diff
@@ -0,0 +1,200 @@
...
+            throw e;
+        }
+
+        // Delegate to global defined converter (e.g. joda or java8)
+        try {
+            ValueExpression ve = clockPicker.getValueExpression("value");
+            if (ve != null) {
+                Class type = ve.getType(context.getELContext());
+                if (type != null && type != Object.class && type != Date.class) {
+                    Converter converter = context.getApplication().createConverter(type);
```
melloware: I think you can do `Converter<Object> converter = ComponentUtils.getConverter(context, component);`

core/src/main/resources/META-INF/primefaces-extensions.taglib.xml
```diff
@@ -10414,5 +10414,68 @@ See showcase for an example. Default null.]]>
...
+            <type>java.lang.String</type>
+        </attribute>
+        <attribute>
+            <description><![CDATA[auto close when minute is selected. Default is false]]></description>
+            <name>autoclose</name>
+            <required>false</required>
+            <type>java.lang.Boolean</type>
+        </attribute>
+        <attribute>
+            <description><![CDATA[	vibrate the device when dragging clock hand. Default is true]]></description>
```
melloware: Remove Extra space after CDATA.

core/src/main/resources/META-INF/resources/primefaces-extensions/clockpicker/1-clockpicker-widget.js
```diff
@@ -0,0 +1,25 @@
+/**
+ * PrimeFaces Extensions ClockPicker Widget.
+ * 
+ */
+PrimeFaces.widget.ExtClockPicker = PrimeFaces.widget.BaseWidget.extend({
```
melloware, line 5: You can just make this `ClockPicker` not `ExtClockPicker` if you want.

core/src/main/resources/META-INF/resources/primefaces-extensions/clockpicker/1-clockpicker-widget.js
```diff
@@ -0,0 +1,25 @@
...
+        this.cfg = cfg;
+        this.jqEl = this.jqId + '_input';
+        this.jq = $(this.jqEl);
+
+        this.jq.clockpicker(this.cfg);
+        // pfs metadata
+        $(this.jqId + '_input').data(PrimeFaces.CLIENT_ID_DATA, this.id);
+        this.originalValue = this.jq.val();
+        
+    }
```
melloware, line 70: We should add `refresh` and `destroy` methods.

We should also exposed `show(), hide(), toggleView()` widgrt methods from https://github.com/weareoutman/clockpicker?tab=readme-ov-file#operations

melloware (changes requested):
See comments

melloware:
Also @jxmai we are doing a 14.0.0-RC1 on Feb 1 if you want to get this in before then!

jxmai:
Great! Thanks for the review. I pushed two commits to address two most important requests. The resource aggregation  and change default type from Object to LocalTime as that's what ClockPicker is specialised for. 

There's still some comments related to the JS api exposure etc (I will review them later and update the PR if needed)

The usage is being simplified because of the value: Object -> LocalTime changes. Originally I thought we might keep the value as Object but that would only raise exceptions since this component does not expect any other value and only would raise exception. Limiting the value option could simplify the code


https://github.com/primefaces-extensions/primefaces-extensions/pull/1424/commits/0bc04a8244e95f438b52267aa6107334e9e3458a#diff-b6211becf538b98657fb8a20e44bb8611fa5717cb97cedb6cb76858aa006a303L182-L197

---

image

jxmai:
TODO note: 

user can type any char in the text box that could break the renderer (this needs to be handled as well).

image

core/src/main/resources/META-INF/primefaces-extensions.taglib.xml
```diff
@@ -10414,5 +10414,68 @@ See showcase for an example. Default null.]]>
...
+            <name>rendered</name>
+            <required>false</required>
+            <type>java.lang.Boolean</type>
+        </attribute>
+        <attribute>
+            <description><![CDATA[Value of the component.]]>
+            </description>
+            <name>value</name>
+            <required>false</required>
+            <type>java.time.LocalTime</type>
```
jepsar: Shouldn't we keep this an Object? You are able to use a converter.

melloware:
I made one fix to your branch. Becuase you were doing `<include>clockpicker/**</include>` that is all you need to minify and compress into the 2 files.  The other stuff was needed if you have files you don't want to minify but yours are straighforward!

core/src/main/resources/META-INF/primefaces-extensions.taglib.xml
jxmai: Yes, you are right. I reverted the change from LocalTime back to Object.

jxmai:
I pushed https://github.com/primefaces-extensions/primefaces-extensions/pull/1424/commits/f7bc85986fd116cde31bc01009ed3ae35104967e to address some issues mentioned by the code review. Please take a look and we can always enhance :handshake:

core/src/main/resources/META-INF/resources/primefaces-extensions/clockpicker/1-clockpicker-widget.js
jxmai: Yes, I checked the rest of the components but they all have Ext as the prefix. Should we keep Ext for consistency purposes?

For example,

https://github.com/primefaces-extensions/primefaces-extensions/blob/927f6b55320df431d91bb9148deac2489adb6c2e/core/src/main/resources/META-INF/resources/primefaces-extensions/inputphone/inputphone-widget.js#L7

core/src/main/resources/META-INF/primefaces-extensions.taglib.xml
jxmai: Resolved via: https://github.com/primefaces-extensions/primefaces-extensions/pull/1424/commits/067ab8c2b57007943f38daf6651a43447f5a45f7#diff-6abac28efd9e878a861db7bf27d79730e05c711fa94cbba609455b44c0766e74L10474

core/src/main/java/org/primefaces/extensions/component/clockpicker/ClockPickerRenderer.java
jxmai: The implementation has been updated. This line does not exist anymore.

core/src/main/java/org/primefaces/extensions/component/clockpicker/ClockPickerRenderer.java
jxmai: The implementation has been updated due to LocalTime update.

core/src/main/java/org/primefaces/extensions/component/clockpicker/ClockPickerRenderer.java
jxmai: Yes, there's no option to input AM/PM in the clock picker component from the usage point of view: https://weareoutman.github.io/clockpicker/. It uses 24-hour format always

core/src/main/resources/META-INF/resources/primefaces-extensions/clockpicker/1-clockpicker-widget.js
jxmai: I included refresh, destroy, show, hide client side API. The toggleView does not work well so I exclude it for now.

melloware (approved):

melloware:
I am doing some refactoring now.

PR #1445, opened 2024-02-14
InputPhone: always submit complete E.164 number

### LINKED ISSUE #1393: InputPhone: always submit complete E.164 number
Add a converter to store phone numbers in E.164 format while keeping validation done in the renderer (if no converter is set).

NicolaIsotta:
fixes #1393 

marking it as draft as I still have to test CSP

### DIFF
--- a/core/src/main/java/org/primefaces/extensions/component/inputphone/InputPhoneRenderer.java
+++ b/core/src/main/java/org/primefaces/extensions/component/inputphone/InputPhoneRenderer.java
@@ -55,7 +55,8 @@ import org.primefaces.util.WidgetBuilder;
 public class InputPhoneRenderer extends InputRenderer {
 
     private static final Logger LOGGER = Logger.getLogger(InputPhoneRenderer.class.getName());
-    private static final String HIDDEN_ID = "_iso2";
+    private static final String HIDDEN_ID = "_hidden";
+    private static final String ISO2_ID = "_iso2";
 
     @Override
     public void decode(final FacesContext context, final UIComponent component) {
@@ -67,7 +68,7 @@ public class InputPhoneRenderer extends InputRenderer {
 
         decodeBehaviors(context, inputPhone);
 
-        final String inputId = inputPhone.getClientId(context) + "_input";
+        final String inputId = inputPhone.getClientId(context) + HIDDEN_ID;
         final String submittedValue = context.getExternalContext().getRequestParameterMap().get(inputId);
 
         if (submittedValue != null) {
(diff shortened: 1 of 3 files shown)


jepsar:
What exactly is the full number? And what is the difference with the number that is posted currently?

NicolaIsotta:
Check https://github.com/jackocnr/intl-tel-input#recommended-usage

> You can always get the full international number (including country code) using `getNumber`, then you only have to store that one string in your database (you don't have to store the country separately), and then the next time you initialise the plugin with that number it will automatically set the country and format it according to the options you specify (e.g. when using `nationalMode` it will automatically display the number in national format, removing the international dial code).

The currently submitted value does not include the international dial code

jepsar:
We just use `nationalMode="false"`, but I guess you want to be using `true` here?

NicolaIsotta:
> We just use `nationalMode="false"`, but I guess you want to be using `true` here?

Yeah, to keep the international prefix in the dropdown, but to have it in the submitted value.

I'm not sure 'bout the attribute name. Maybe `fullValue` is not clear enough? I'm thinking about `fullValueSubmit`/`fullNumberSubmit` or `submitFullValue`/`submitFullNumber`

@melloware I targeted 13.X as it's not a breaking change, let me know if have to change to 14.X only

melloware:
13.x is fine.  i am not sure about fullValue either maybe "rawValue" or "completeValue" ?

jepsar:
I'm not sure whether to make this an option. Also, I think I just might want to make E.164 the default and allow users to convert from there (if needed). If currently incomplete numbers are returned in certain scenarios, that's a bug IMO.

NicolaIsotta:
- Now the submitted value is always E.164-formatted
- The utils.js is required by the `getNumber()` function
- tested with CSP

core/src/main/java/org/primefaces/extensions/component/inputphone/InputPhoneRenderer.java
```diff
@@ -204,13 +206,11 @@ protected void encodeScript(final FacesContext context, final InputPhone inputPh
...
         encodeCountries(wb, "preferredCountries", inputPhone.getPreferredCountries());
 
-        if (inputPhone.isUtilsScriptRequired()) {
-            wb.attr("utilsScript",
-                        context.getApplication()
-                                    .getResourceHandler()
-                                    .createResource("inputphone/utils.js", "primefaces-extensions")
-                                    .getRequestPath());
-        }
+        wb.attr("utilsScript",
```
jepsar, line 209: As this is now always loaded, I think we can simply aggregate it into the JavaScript for this component in the pom.xml.

https://github.com/primefaces-extensions/primefaces-extensions/blob/4150d88055e475772564acf29456a4d9f68c07a0/core/pom.xml#L247-L259

jepsar: Never mind. The library needs a URL and will load it.

melloware:
Is this PR ready?

jepsar:
I would like to test this. Maybe I can do it today.

jepsar (approved):

jepsar:
@NicolaIsotta thanks for your fix!

PR #1455, opened 2024-02-26
new InputOtp component

### LINKED ISSUE #1453: New component: OTP / PIN input
We need a OTP input for a new project, I checked primefaces and extensions' showcase but I haven't found anything suitable.

Some examples found in the web:

- https://www.geeksforgeeks.org/create-otp-input-field-using-html-css-and-javascript/
- https://shuqikhor.github.io/Vanilla-OTP-Input/
- https://foolishdeveloper.com/otp-input-field-javascript/
- https://github.com/devfolioco/react-otp-input
- https://dominicarrojado.com/posts/how-to-create-your-own-otp-input-in-react-and-typescript-with-tests-part-1/

I may try to put something together if you guys think this could be useful.

NicolaIsotta:
fixes #1453

Looks nice so far. Still TODO:

- [x] decide the final name. Alternatives: inputPin, inputOTP, pin, otp
- [x] cleanup taglib attributes
- [x] add utility method to widget (eg. set/getValue)
- [x] server-side validation for `numeric="true"`
- [x] aria-label
- [x] optional separator
- [x] placeholder
- [x] RTL
- [x] align with [PrimeVue InputOtp](https://primevue.org/inputotp)


showcase/src/main/webapp/sections/inputpin/basicUsage.xhtml
```diff
@@ -0,0 +1,37 @@
...
+      xmlns:f="http://xmlns.jcp.org/jsf/core"
+      xmlns:ui="http://xmlns.jcp.org/jsf/facelets"
+      xmlns:showcase="http://primefaces.org/ui/extensions/showcase">
+    <ui:composition template="/templates/showcaseLayout.xhtml">
+        <ui:define name="centerContent">
+            <f:facet name="header">
+                <h:outputText value="InputPin"/>
+            </f:facet>
+            <h:panelGroup layout="block" styleClass="centerContent">
+                International phone numbers input.
```
jepsar: Please change

core/src/main/java/org/primefaces/extensions/component/inputpin/InputPinRenderer.java
```diff
@@ -0,0 +1,164 @@
...
+        final String inputStyleClass = getStyleClassBuilder(context)
+                    .add(InputText.STYLE_CLASS, InputPin.CELL_STYLE_CLASS)
+                    .add(inputPin.getInputStyleClass())
+                    .build();
+        final char[] chars = valueToRender.toCharArray();
+        for (int i = 1; i <= inputPin.getSize(); i++) {
+            final String inputId = clientId + InputPin.INPUT_SUFFIX + i;
+            final String inputValue = chars.length >= i ? String.valueOf(chars[i - 1]) : "";
+
+            writer.startElement("input", null);
```
jepsar: Do we need maxlength 1 here?

jepsar (changes requested):

core/src/main/java/org/primefaces/extensions/component/inputpin/InputPinRenderer.java
NicolaIsotta: with maxlength="1" it wouldn't be possible to paste longer text (AFAIK)

core/pom.xml
```diff
@@ -270,6 +270,30 @@
...
+                        <resourcesSet>
+                            <includes>
+                                <include>inputpin/inputpin.css</include>
+                            </includes>
+                            <aggregations>
+                                <aggregation>
+                                    <inputDir>${resources.dir.compressed}</inputDir>
+                                    <outputFile>${resources.dir.compressed}/inputpin/inputpin.css</outputFile>
+                                    <withoutCompress>${withoutCompress}</withoutCompress>
+                                </aggregation>
```
melloware: You can change this section to be in the above `<include>inputpin/**</include>` and it will compress the JS and CSS and name them `inputpin.js` and `inputpin.css`.

core/src/main/resources/META-INF/resources/primefaces-extensions/inputpin/inputpin-widget.js
```diff
@@ -0,0 +1,198 @@
...
+                }
+                $this.updateInput();
+
+                if (originalOninput && originalOninput.call(this, e) === false) {
+                    return false;
+                }
+            });
+
+            $(input).prop('onkeydown', null).off('keydown').on('keydown', function (e) {
+                switch (e.keyCode) {
```
melloware:
```suggestion
                switch (e.code) {
```

core/src/main/resources/META-INF/resources/primefaces-extensions/inputpin/inputpin-widget.js
```diff
@@ -0,0 +1,198 @@
...
+                    case 8: // backspace button
```
melloware:
```suggestion
                    case 'Backspace':
```

core/src/main/resources/META-INF/resources/primefaces-extensions/inputpin/inputpin-widget.js
```diff
@@ -0,0 +1,198 @@
...
+                            }
+
+                            // clear previous box and focus on it
+                            inputsJq[i - 1].value = '';
+                            inputsJq[i - 1].focus();
+                            inputsJq[i - 1].dispatchEvent(new Event("input"));
+                            return;
+                        }
+                        break;
+                    case 46: // delete button
```
melloware:
```suggestion
                    case 'Delete':
```

core/src/main/resources/META-INF/resources/primefaces-extensions/inputpin/inputpin-widget.js
```diff
@@ -0,0 +1,198 @@
...
+
+                            // clear the last box
+                            inputsJq[inputsJq.length - 1].value = '';
+                            inputsJq[i].select();
+                            e.preventDefault();
+                            inputsJq[inputsJq.length - 1].dispatchEvent(new Event("input"));
+                            return;
+                        }
+                        break;
+                    case 37: // left button
```
melloware:
```suggestion
                    case 'ArrowLeft`:
```

core/src/main/resources/META-INF/resources/primefaces-extensions/inputpin/inputpin-widget.js
```diff
@@ -0,0 +1,198 @@
...
+                        if (i > 0) {
+                            e.preventDefault();
+                            inputsJq[i - 1].focus();
+                            inputsJq[i - 1].select();
+                        }
+                        return;
+                    case 39: // right button
```
melloware:
```suggestion
                    case 'ArrowRight':
```

core/src/main/resources/META-INF/resources/primefaces-extensions/inputpin/inputpin-widget.js
```diff
@@ -0,0 +1,198 @@
...
+
+        // style disabled if necessary
+        if (this.disabled) {
+            PrimeFaces.utils.disableInputWidget(this.inputsJq);
+            this.hinput.attr("disabled", "disabled");
+        }
+
+        // visual effects
+        PrimeFaces.skinInput(this.inputsJq);
+
```
melloware: Add ARIA Label to inputs something like...

```
.attr('aria-label', `Please enter OTP character ${index + 1}`,
```

melloware (changes requested):

melloware:
Do a git pull i just merged my changes into your branch.

core/src/main/java/org/primefaces/extensions/component/inputpin/InputPinRenderer.java
```diff
@@ -0,0 +1,169 @@
...
+
+            if ("password".equalsIgnoreCase(inputPin.getType())) {
+                writer.writeAttribute("type", inputPin.getType(), null);
+            }
+
+            if (inputPin.isNumeric()) {
+                writer.writeAttribute("inputmode", "numeric", null);
+            }
+
+            writer.writeAttribute(HTML.ARIA_LABEL, MessageFactory.getMessageString(context, MESSAGE_ARIA_LABEL_KEY, i), null);
```
melloware: I know this is easier here but we have been trying to move all ARIA Labels to the client side in JS code because it is easier to update for user and cleaner.  The mix of server side and client side labels was always a bad idea and in 14 we have moved "most" of the labels to the client side.

melloware (changes requested):

core/src/main/java/org/primefaces/extensions/component/inputpin/InputPinRenderer.java
NicolaIsotta: ok, I will change

NicolaIsotta: Is there a localization file for js in the extensions project?

melloware: right now there is not so what I would do is do something like this in the init.

```js
this.ariaLabel = this.cfg.ariaLabel|| PrimeFaces.getAriaLabel('inputpin.LABEL');
``` 

That way in the future if we add it to PF core translations or people can override.

jepsar:
> decide the final name. Alternatives: inputPin, inputOTP, pin, otp

I like inputPin the most.

melloware:
+1 for inputPin as well.

NicolaIsotta:
Not sure about RTL, I don't use it and I don't really know how to implement it. There aren't  many examples on the web...

jepsar:
> Not sure about RTL, I don't use it and I don't really know how to implement it. There aren't many examples on the web...

Haven't looked at it in detail, but can't we simply use CSS for it?

Normal:
```css
    display: flex;
```
RTL, add:
```css
    flex-direction: row-reverse;
```

NicolaIsotta:
> > Not sure about RTL, I don't use it and I don't really know how to implement it. There aren't many examples on the web...
> 
> Haven't looked at it in detail, but can't we simply use CSS for it?
> 
> Normal:
> 
> ```css
>     display: flex;
> ```
> 
> RTL, add:
> 
> ```css
>     flex-direction: row-reverse;
> ```

Had to tweak a bit but it works!

BTW, it think the component it's almost ready, I want to test a little bit more and then I'll mark the PR as ready

melloware:
awesome yeah that RTL looks good!

NicolaIsotta:
I think the component's ready.

BTW, I've seen PrimeVue now includes an [InputOtp](https://primevue.org/inputotp) component. Should we align component and attributes' names?

melloware:
Yes we should because this component may eventually move into PrimeFaces core! So we should keep it aligned as InputOtp.

Also can you review their keyboard handling just to make sure we are in alignment?  Just making sure we are doing the same as they are: https://github.com/primefaces/primevue/blob/master/components/lib/inputotp/InputOtp.vue

Also it looks like they are not using ARIA label so the good news is I will probably be able to get this label added to PrimeLocale since PrimeVue will also need it!

NicolaIsotta:
I'll check and align.

BTW, I already discovered a bug, they're using wrong input mode.

melloware:
HA yeah and I just added this PR to PrimeLocale: https://github.com/primefaces-extensions/primefaces-extensions/issues/1453

melloware:
@NicolaIsotta i am going to merge this so I can publish 14.0.0-RC2.  We can continue tweaking before 14.0.0-Final.

melloware (approved):

NicolaIsotta:
I was just about to mark this as ready to merge, as far as I can tell the component is ok and is almost aligned with PrimeVue - the only thing missing is templating but it's (partially) replaced by the "separator" attribute.

melloware:
OK i am going to open a new ticket

melloware:
@NicolaIsotta i made a few more tweaks from PrimeVue and also changed 1 thing.

See: https://github.com/primefaces-extensions/primefaces-extensions/commit/79b8b22eb9a2235836c6bb80ad9846d9eb064b94

I made Delete and BackSpace do the same thing as to me it was the behavior I was expecting when pressing the delete key.  Thoughts?

jepsar:
As we are aligning with a Vue component, shouldn't we move this to PrimeFaces 15? Shall I create an issue for it?

melloware:
I totally agree @jepsar this should be moved to core PF in 15 to align with other PrimeTek components.

jepsar:
I think we should keep it here in 14 for public beta testing and then promote it to PF in 15.

melloware:
Oh yes absolutely we release here in 14 so people can test and report any issues so it will be rock solid for porting in 15..

I would actually like to do the same with InputPhone and InputPlace but i will post something on the team Discord about it.

NicolaIsotta:
> @NicolaIsotta i made a few more tweaks from PrimeVue and also changed 1 thing.
> 
> See: [79b8b22](https://github.com/primefaces-extensions/primefaces-extensions/commit/79b8b22eb9a2235836c6bb80ad9846d9eb064b94)
> 
> I made Delete and BackSpace do the same thing as to me it was the behavior I was expecting when pressing the delete key. Thoughts?

TBH I prefer Vanilla-OTP-Input behavior, but it's just personal taste. I'm fine with the modifications.

BTW I haven't looked at the template feature yet because I'm pretty busy at the moment, but I'll try to work on it ASAP

melloware:
Thanks @NicolaIsotta !  Yeah the Delete key just felt like in my brain more natural do be treated the same as backspace.  if people complain we can always update but i have a feeling no one will!

BTW your version with our tweaks actually works better than the PrimeVue one.  I don't like how they handled "paste" vs how you did it. Yours is cleaner and works great.

melloware:
Also just made one more fix we were not forcing the required message if a user filled out 3 of 4 digits.  if its a 4 digit code then all 4 digits should be required.

PR #1456, opened 2024-02-27
Fix #1321 - MorphiaLazyModel: Refactor in 14.0.0

### LINKED ISSUE #1321: MorphiaLazyModel: Refactor in 14.0.0
Based on core PF changes: https://github.com/primefaces/primefaces/pull/10776

cc @Rapster cc @newk5

Rapster:
@newk5 for review

### DIFF
--- a/core/src/main/java/org/primefaces/extensions/model/mongo/MorphiaLazyDataModel.java
+++ b/core/src/main/java/org/primefaces/extensions/model/mongo/MorphiaLazyDataModel.java
@@ -21,6 +21,22 @@
  */
 package org.primefaces.extensions.model.mongo;
 
+import java.io.Serializable;
+import java.lang.reflect.Field;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.function.BiConsumer;
+import java.util.function.Consumer;
+import java.util.logging.Logger;
+
+import javax.faces.FacesException;
+import javax.faces.context.FacesContext;
+import javax.faces.convert.Converter;
+
 import dev.morphia.Datastore;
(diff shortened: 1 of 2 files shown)


newk5:
Thanks! Sorry I haven't been able to look into this yet, I'm not familiar with the new LazyDatamodel refactor that was done recently in PF. 

I was now looking through this commit and I agree with all changes except one in particular with the builder. I think we should rename the entityManager(etc..) to datastore(etc..) because morphia has no concept of entity managers, entity managers usually mean JPA related stuff. I understand this probably came from the JPALazyDataModel in PF, and wanting to normalize the API to be the same between the JPALazyModel and MorphiaLazyDataModel but for morphia users, the term "entityManager" will probably cause confusion, since morphia always refers to it as "datastore"

Rapster:
Yup, just an oversight

core/src/main/java/org/primefaces/extensions/model/mongo/MorphiaLazyDataModel.java
```diff
@@ -344,41 +325,93 @@ protected Method getRowKeyGetter() {
...
+        public Builder() {
+            model = new MorphiaLazyDataModel<>();
+        }
+
+        public Builder<T> entityClass(Class<T> entityClass) {
+            model.entityClass = entityClass;
+            return this;
+        }
+
+        public Builder<T> entityManager(SerializableSupplier<Datastore> datastore) {
```
melloware: EntityManager to Datastore and then I think this is great!

melloware (changes requested):

melloware (approved):

Rapster:
Be aware I couldn't test it (or we have a branch for that?)

melloware:
We don't. This code was donated by @newk5 as he uses it heavily in his company.

PR #1460, opened 2024-03-02
Fix #1459: SunEditor prevent HTML manipulation

### LINKED ISSUE #1459: SunEditor: Allow Untouched HTML Option
### Describe the bug

Apparently this can be done with...

```js
editorInstance.core.cleanHTML = function (html) {return html;}
editorInstance.core.convertContentsForEditor = function (html) {return html;}
```

### Reproducer

_No response_

### Expected behavior

_No response_

13.0.7

### JSF implementation

All

### JSF version

2.3

### Browser(s)

_No response_

melloware:
Fix #1459: SunEditor prevent HTML manipulation

cc @jepsar @mkomko @stolp


jepsar:
Was thinking whether to make this optional, but I guess most Faces users would want to clean-up the HTML at the server side.

melloware:
That is what I was thinking too.  He really shouldn't be touching the HTML in our case we can choose to do what we want with it.

PR #1479, opened 2024-03-12
Fix #137: InputPlace component

### LINKED ISSUE #137: New Component: InputPlace
Add google place autocomplete component.

https://developers.google.com/maps/documentation/javascript/examples/places-autocomplete

Since we already have Geocoding and Rev. Geocoding from Google, we should also have other google maps features like Place autocomplete etc.

### LINKED ISSUE #1482: InputPhone: Not respecting inputStyleClass
### Describe the bug

using `inputStyleClass="test"` you will not see test appear on the input.

### Reproducer

_No response_

### Expected behavior

class should be applied

13.0.7

### JSF implementation

All

### JSF version

2.3

### Browser(s)

_No response_

melloware:
Fix #137: InputPlace component
Fix #1482 

image

cc @NicolaIsotta


melloware:
I am going to add another example linking the InputPlace to GMap.

melloware:
Need to add long and short values for country and possible other fields like name.

core/src/main/java/org/primefaces/extensions/component/inputplace/InputPlaceRenderer.java
```diff
@@ -0,0 +1,113 @@
...
+
+        decodeBehaviors(context, inputPlace);
+
+        String clientId = inputPlace.getClientId(context);
+        String submittedValue = context.getExternalContext().getRequestParameterMap().get(clientId);
+
+        if (submittedValue != null) {
+            int maxlength = inputPlace.getMaxlength();
+            if (maxlength > 0 && submittedValue.length() > maxlength) {
+                submittedValue = LangUtils.substring(submittedValue, 0, maxlength);
```
jepsar, line 59: This is kind of weird. Maybe we should log a warning here so the user can consider using a larger maximum length in their model.

core/src/main/java/org/primefaces/extensions/model/inputplace/Place.java
```diff
@@ -0,0 +1,174 @@
...
+        this.country = params.get(clientId + "_country");
+        this.latitude = Double.parseDouble(params.get(clientId + "_lat"));
+        this.longitude = Double.parseDouble(params.get(clientId + "_lng"));
+        this.administrativeAreaLevel1 = params.get(clientId + "_administrative_area_level_1");
+        this.administrativeAreaLevel2 = params.get(clientId + "_administrative_area_level_2");
+        this.administrativeAreaLevel3 = params.get(clientId + "_administrative_area_level_3");
+        this.types = params.get(clientId + "_types");
+    }
+
+    public Place(String placeId, String formattedAddress, String addressLine, String postalCode, String city, String state, String country,
```
jepsar: Wow! Sonar does not like when the amount of parameters exceeds 7, and neither do I 🙈. Maybe we could have a builder instead?

jepsar (changes requested):
I found two small things to consider.

core/src/main/java/org/primefaces/extensions/model/inputplace/Place.java
jepsar: Is this even used? (Just did a rough scan). Maybe we an simply delete it. Or make the fields non-final.

core/src/main/java/org/primefaces/extensions/component/inputplace/InputPlaceRenderer.java
melloware: This is copied directly from InputTextRenderer.java in PF?

core/src/main/java/org/primefaces/extensions/model/inputplace/Place.java
melloware: Agreed I was just following your pattern but I will remove the constructor!

core/src/main/java/org/primefaces/extensions/component/inputplace/InputPlaceRenderer.java
jepsar: Ah, okay, was thinking it might have been that Google's result was too large for the field's maxlength. But I guess that's clipped in the UI / JS.

jepsar (dismissed):

melloware:
@jepsar working on a custom Address Entry form using InputPlace + GMap + InputPhone.  

I automatically set the InputPhone `initialCountry="#{testView.selectedPlace.countryCode}"` to the Google Place returned Country Code!

image

I did find one minor issue with InputPhone when trying to use a Responsive form.  It won't size the field even with `w-full`

```xml
<div class="field col-12 md:col-6">
    <p:outputLabel value="Primary contact phone" for="@next" />
    <pe:inputPhone id="txtPhone" initialCountry="#{testView.selectedPlace.countryCode}" value="#{testView.phone}" required="true"
        styleClass="w-full" inputStyleClass="w-full" formatOnDisplay="true" />
</div>
```

Its because the outer `span` of course ignores size because its a span and the inner Input `w-full` expands to the width of the ITI Div.  So the only way I could get it to behave the way I wanted was to add CSS...

```css
/* InputPhone Full Width */
.iti {
    width: 100%;
}
```

PR #1481, opened 2024-03-13
InputPhone: wrap input event

NicolaIsotta:
Without wrapping, this is what happens with a `p:ajax event="input"`
inputhphonr
(submitted value is one step behind)

### DIFF
--- a/core/src/main/resources/META-INF/resources/primefaces-extensions/inputphone/inputphone-widget.js
+++ b/core/src/main/resources/META-INF/resources/primefaces-extensions/inputphone/inputphone-widget.js
@@ -67,8 +67,23 @@ PrimeFaces.widget.ExtInputPhone = PrimeFaces.widget.BaseWidget.extend({
                 $this.callBehavior('countrySelect', ext);
             }
         });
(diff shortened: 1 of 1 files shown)


melloware (approved):

melloware:
I will backport this to 13.0.8

melloware:
Actually I can't because inputphone has so many changes and improvements since the 13.x branch i would have to backport the entire InputPhone but i don't think @jepsar wanted me to do that!

NicolaIsotta:
This is not needed in 13.x, it's an improvement of #1393 implementation which has been ported to 14.x only

jepsar:
> Actually I can't because inputphone has so many changes and improvements since the 13.x branch i would have to backport the entire InputPhone but i don't think @jepsar wanted me to do that!

I personally don't mind, but this PR can be a breaker for people https://github.com/primefaces-extensions/primefaces-extensions/pull/1445

melloware:
OK yeah lets leave everything in 14.X as i hate introducing breaking change in a minor patch release.

PR #1490, opened 2024-03-18
Fix #1484: Markdown Editor

### LINKED ISSUE #1484: New Component: Markdown Editor
Can anyone recommend a Markdown editor which uses a toolbar (like on GitHub comments for example)?

 I've found  https://github.com/Ionaru/easy-markdown-editor.

Cc @jepsar

melloware:
Fix #1484: Markdown Editor


core/src/main/java/org/primefaces/extensions/component/base/AbstractEditorInputTextArea.java
```diff
@@ -0,0 +1,195 @@
...
+        getStateHelper().put(PropertyKeys.toolbar, toolbar);
+    }
+
+    /**
+     * Enforce security by default requiring the OWASP sanitizer on the classpath. Only if a user marks the editor with secure="false" will they opt-out of
+     * security.
+     *
+     * @param context the FacesContext
+     */
+    public void checkSecurity(FacesContext context) {
```
jepsar, line 162: Markdown != HTML. Maybe keep this in the SunEditor?

melloware: But in Markdown you can put raw HTML like <b> and HTML Links etc and Media Images etc.  I thought it was safer to allow sanitization?

There is a lot of discussion about XSS See: https://github.com/Ionaru/easy-markdown-editor/issues/64

jepsar (approved):

PR #1541, opened 2024-05-01
#1540 : Allow custom morphia FindOptions/CountOptions

### LINKED ISSUE #1540: MorphiaLazyDataModel: Allow user to specify FindOptions and CountOptions

When using the MorphiaLazyDataModel we should allow the user to specify their own FindOptions and CountOptions. These are morphia objects that allow customizing the query even further like specifying different collections to query against rather than the default one specified on the class, sorting by specific fields and more.

We added this to our own MorphiaLazyDataModel at work since we needed to point to a different collection in some cases but still use the same model class. We also wanted to be able to sort by a specific field not visible as column on the p:datatable

Will provide a PR for this

_No response_

### Additional context

_No response_

newk5:
Fix  #1540

### DIFF
--- a/core/src/main/java/org/primefaces/extensions/model/mongo/MorphiaLazyDataModel.java
+++ b/core/src/main/java/org/primefaces/extensions/model/mongo/MorphiaLazyDataModel.java
@@ -31,6 +31,7 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.function.BiConsumer;
 import java.util.function.Consumer;
+import java.util.function.Supplier;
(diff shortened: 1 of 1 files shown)


melloware (approved):

melloware:
Ahhh crap this will have to make 14.0.1!

newk5:
> Ahhh crap this will have to make 14.0.1!

Ahah yea I just noticed I missed the 14.0.0 release by a few minutes. It's ok, I'll wait for 14.0.1

Issue #128, opened 2019-09-28
New Component: Mapping OL3 or OpenStreetMap


melloware:
Based on 2 requests in PF...

https://github.com/primefaces/primefaces/issues/67
https://github.com/primefaces/primefaces/issues/40

OL3:  https://openlayers.org/

OpenSteetMap: https://www.openstreetmap.org/#map=5/38.007/-95.844

melloware:
OK added to 14.0.1 and you can use the same Java model objects as are in PF core for GMAP. @mores 

I added 1 basic showcase example but @mores we probably need a few more like Markers etc.

mores:
> @mores we probably need a few more like Markers etc.

I created an "Event Usage"  - https://github.com/mores/primefaces-extensions/tree/moreExamples

I will continue to build out more examples.

melloware:
Excellent!

PR #1562, opened 2024-06-01
[OpenStreetMap] Fixes dragend event / Adds more examples

mores:
https://github.com/primefaces-extensions/primefaces-extensions/issues/128

Examples show:
Events
Markers
Marker Selection
Draggable Markers
Shapes


melloware (approved):
Thanks!

PR #1640, opened 2024-08-15
Fix #1639: SunEditor do not escape value

### LINKED ISSUE #1639: SunEditor C Code loses <stdio.h> - strictMode does not apply
### Describe the bug

Version Used: 14.0.4
- In the sunEditor, i enter the following C code (choosing the code in the drop down). The code is saved in the database after the < being converted to ampersand lt ;
`#include<stdio.h>`

int main()
{
   //more code here
}

- Later when I edit the content again, <stdio.h> is lost and only #include is present in the editor.

### Reproducer

_No response_

### Expected behavior

The <stdio.h> must be retained.

14.0.4

### JSF implementation

Mojarra

### JSF version

3.0.1

### Browser(s)

Chrome, Firefox

melloware:
Fix #1639: SunEditor do not escape value

I looked at what PF InputTextArea is doing and it does not escape the output.


bercolax:
I modified the code in SunEditorRenderer.java, built using Maven and the fix works. Thanks for your time involved.

melloware:
No problem this will be in 14.0.5

PR #1741, opened 2024-11-02
Upgrade Reveal.js to 5.1.0

### LINKED ISSUE #1485: KeyNote: Update Reveal.js
@aripddev can you look at upgrading to the latest Reveal.js for KeyNote component?

Right now its the only component that doesn't work for CSP so i was hoping their newer version fixed that.

aripddev:
@melloware Sorry for the late reply. I just noticed your ping. Here is the PR for the issue #1485 .

Fix #1485


melloware:
Good new I confirmed this allows PrimeFaces.CSP to be enabled now!

PR #1754, opened 2024-11-11
PrimeFacesResourceProcessor check if JS/CSS already loaded

melloware:
Related to: https://github.com/primefaces/primefaces/issues/12855


melloware:
cc @blutorange

PR #1757, opened 2024-11-13
Fix #1756: EscapeSanitizerComponentListener

### LINKED ISSUE #1756: EscapeSanitizerComponentListener: New Component

Request from a client they automatically wants all `escape="false"` components to go through the HtmlSanitizer so that way `bold` will work but `<script>` etc will get stripped.

Yes customer had `<h:outputText escape="false" >` but it was from User input and Pen tester was able to add XSS.  The developer didn't realize what escape="false" was for so it was a security hole they didn't even know.   Customer has 70+ apps and wants a global solution.

```java
public class EscapeSanitizerComponentListener implements SystemEventListener {
```

melloware:
Fix #1756: EscapeSanitizerComponentListener

### DIFF
--- a/core/src/main/java/org/primefaces/extensions/application/EscapeSanitizerComponentListener.java
+++ b/core/src/main/java/org/primefaces/extensions/application/EscapeSanitizerComponentListener.java
@@ -1,0 +1,119 @@
+/*
+ * Copyright (c) 2011-2024 PrimeFaces Extensions
+ *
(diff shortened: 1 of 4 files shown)


melloware:
Thanks to @FlipWarthog for this idea

FlipWarthog (approved):

jepsar:
Oh, nice!

PR #1768, opened 2024-11-20
Migrate combinedResourceHandler.xhtml to jakarta

mkomko:


melloware:
I was going to do this with 15.0.0 when we change the whole showcase to Jakarta.

PR #1795, opened 2024-12-08
Fixes #1794

### LINKED ISSUE #1794: Component: DynaForm "Advanced features" showcase broken since 14.0.7 (Button does not submit via AJAX)
### Describe the bug

The "advanced features" tab of the DynaForm showcase seems to be broken since 14.0.7. Not sure what the issue is exactly. It's working fine locally with the `14.0.6` tag and below, and doesn't work with `14.0.7` and `14.0.8`. At first glance, I didn't see any change that might be responsible, though.

### Reproducer

* Pull the current master branch and start the showcase
* Go to http://localhost:8080/showcase-ext/sections/dynaform/advancedFeatures.jsf
* After 1-2 seconds, an error is logged to the console (`Widget for var 'dynaFormWidget' not available!`) and the page becomes blank

The JS error seems unrelated -- the widget calls `PF(...)` to check if the widget exists and that just logs an error message.

What seems to be happening is that the showcase set the `autoSubmit` property on the DynaForm to `true`. ~~The client side JavaScript then submits the form. The server, when it processes that POST request, notices a CSP nonce mismatch and throws an exception. This results in an unfinished response, which probably explains the blank page.~~ The main issue seems to be that the submit is not made via AJAX anymore, see comment below.

Java logs a CSP issue:
 
```
org.primefaces.csp.CspException: CSP nonce mismatch
	at org.primefaces.csp.CspState.getNonce(CspState.java:90)
	at org.primefaces.csp.CspPhaseListener.initCsp(CspPhaseListener.java:96)
	at org.primefaces.csp.CspPhaseListener.beforePhase(CspPhaseListener.java:68)
	at org.apache.myfaces.lifecycle.PhaseListenerManager.informPhaseListenersBefore(PhaseListenerManager.java:77)
	at org.apache.myfaces.lifecycle.LifecycleImpl.render(LifecycleImpl.java:259)
	at javax.faces.webapp.FacesServlet.service(FacesServlet.java:206)
	at org.eclipse.jetty.servlet.ServletHolder$NotAsync.service(ServletHolder.java:1459)
```

### Expected behavior

Behaves as befores: It submits the form and displays the result in a dialog.

14.0.7

### JSF implementation

None

### JSF version

_No response_

### Browser(s)

_No response_

blutorange:
Fixes #1794. Just delaying the submission seems to work well.

Delay submission until the DOM is ready. When CSP is enabled, the AJAX click listener for the button is only registered by a script at the end of the <body>.

### DIFF
--- a/core/src/main/resources/META-INF/resources/primefaces-extensions/dynaform/0-dynaform.js
+++ b/core/src/main/resources/META-INF/resources/primefaces-extensions/dynaform/0-dynaform.js
@@ -17,7 +17,9 @@ PrimeFaces.widget.ExtDynaForm = PrimeFaces.widget.BaseWidget.extend({
         }
 
         if (cfg.autoSubmit && !PF(cfg.widgetVar)) {
-            this.submitForm();
+            // Delay submission until the CSP handler was registered, needed when CSP is enabled
+            // See https://github.com/primefaces-extensions/primefaces-extensions/issues/1794#issuecomment-2526318388
+            $(() => this.submitForm());
         } else if (cfg.isPostback && this.toggledExtended && this.uuid == cfg.uuid) {
             var rows = this.jq.find("tr.pe-dynaform-extendedrow");
             if (rows.length > 0) {


blutorange:
Seems to have been a temporary NPM failure, pipeline works now.

melloware (approved):

Issue #1864, opened 2025-02-26
DocumentViewer: Upgrade pdf.js version to v4.10.38


Teutron-flo:
### Describe the bug

Hi all,

after the update from pf-extensions 13.0.5 to 15.0.0, I got a couple of issues while displaying our pdfs.
I can see that with https://github.com/primefaces-extensions/primefaces-extensions/issues/1569 
for version pf-extensions 13.0.11, the pdf.js got a version update to pdf.js-4.3.136
https://github.com/primefaces-extensions/primefaces-extensions/pull/1574

So one of the changes, within the update creates now the issues on our side.
Thats why I wanted to ask, if it is possible to get maybe again an update the the latest version?
https://github.com/mozilla/pdf.js -> https://github.com/mozilla/pdf.js/releases/tag/v4.10.38

And I hope, I get out display issues solved afterwards.

BR
Florian

### PrimeFaces Extensions version

15.0.0

### JSF implementation

None

melloware:
Can you explain what your display issues are?

Teutron-flo:
-- Image deleted for the moment

On the right side the screenshot of the webpage / leftside the downloaded pdf.
Inside the browser log you can see a couple of issues with loading of the fonts.
If i replace pe:docmentViewer with p:media it works also, but p:media has some gernal issues with different mobile devices.

melloware:
Are there any errors in the F12 console?  to me it looks like fonts are not being loaded properly.

can you try `disableFontFace="false"` ?

Teutron-flo:
As I said there are some font errors inside the Browser log like:
Warning: getPathGenerator - ignoring character: "Error: Requesting object that isn't resolved yet Sans_path_Z.".

When I try it with disableFontFace="false", it get even more bad:

-- Image deleted for the moment

melloware:
OK i just wanted to confirm.  OK so its that your custom fonts inside the PDF are not loading properly.

melloware:
Is there any way you could share a redacted version of the PDF with me so I can try and debug?

Teutron-flo:
You should have received a mail under "mellowaredev at gmail com".

melloware:
Got it.

melloware:
Seems like this issue: https://github.com/mozilla/pdf.js/issues/19234

melloware:
[...]

PR #1897, opened 2025-03-25
Update pdf.viewer.js

Teutron-flo:
PrimeFaces (#1864), bugfix when you have embedded and not embedded fonts inside the same pdf. Normally pdf.js is able to handle this, but not in our case with the displaying iframe.

The orginal change / idea for the change was not from me, i found it here https://github.com/mozilla/pdf.js/issues/11311#issuecomment-863714376


melloware:
This is great!

melloware:
I will backport this to 14.X as well.

melloware:
15.0.3 is now in Maven Central with this fix!

PR #1917, opened 2025-04-07
#1916 Layout: self collapse when open with double mouse click - fix

### LINKED ISSUE #1916: Layout: self collapse when open with double mouse click
### Describe the bug

 As shown in the attached recording, when I double-click outside the button and then move the mouse outside the panel, it collapses. Do you have any idea why this is happening?

https://github.com/user-attachments/assets/f50e9093-9ee4-460f-9fe0-7d49f9ff3a19

### Reproducer

_No response_

### Expected behavior

_No response_

Lastest - showcase 

### JSF implementation

None

### JSF version

As in showcase

### Browser(s)

_No response_

kamilkrzywanski:
Fix #1916

### DIFF
--- a/core/src/main/resources/META-INF/resources/primefaces-extensions/layout/0-jquery.layout.js
+++ b/core/src/main/resources/META-INF/resources/primefaces-extensions/layout/0-jquery.layout.js
@@ -3931,7 +3931,7 @@
 
                         $R
                             // add or remove event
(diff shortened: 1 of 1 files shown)


melloware:
#Hmm. Would it be better to just mark your panel as `slidable=false` to disable the sliding entirely?

PR #1991, opened 2025-06-06
Fix #1990: Clipboard add DOM element

### LINKED ISSUE #1990: Clipboard: Destroyed detached widget: ...
i always get: Destroyed detached widget: widget_formPositionen_j_id_4o_0_etikett_clip

it seems that clipboard does not render a dummy markup, like other updateable scripts are doing

melloware:
Fix #1990: Clipboard add DOM element

@tandraschko maybe there is something I am missing but even adding this I am still seeing the `Destroyed detached widget: widget_clipCopy` and I don't understand why.


JuanMiguelBG-Ticarum:
This change breaks v13.0 branch.

The function `_remove` doesn't exist in that branch.

melloware:
Thanks i would stay on 13.0.19 for now then...

melloware:
@JuanMiguelBG-Ticarum i reverted that and pushing 13.0.21 now.

PR #2101, opened 2025-09-16
Added fullscreen support for Osmap

kurtomerfaruk:


melloware:
@kurtomerfaruk great addition!  Can you run `mvn clean package` and inspect the built JAR look in META-INF and see if the leaflet.js was combined into one?  if so then the `@Resources` at the top of the java file will be wrong in production mode.

kurtomerfaruk:
I see that the leaflet files within the jar file have not been merged.

melloware:
Thank you for the PR!

melloware:
i will put this in 15.0.8

PR #2207, opened 2025-12-11
Fix #2202 : no role=row/gridcell when role=presentation is specified

### LINKED ISSUE #2202: DynaForm : Hardcoded ARIA roles (row/gridcell) prevent using role="presentation" for layout tables (WCAG violation)
### Describe the bug

I am using pe:dynaForm (version 15.0.10) to generate a form layout. 
Since this is purely a layout table for form inputs and not a data grid, I am attempting to adhere to WCAG best practices by adding role="presentation" (via pass-through attributes) to the valid parent container to hide the table semantics from screen readers.
However, pe:dynaForm automatically renders role="row" on <tr> elements and role="gridcell" on <td> elements. These explicit attributes cause an accessibility violation when the parent table has role="presentation", because elements with role row require a semantic parent of role grid, treegrid, or table.

References: 
[WCAG 4.1.2](https://www.w3.org/WAI/WCAG21/Understanding/name-role-value.html) - Name, Role, Value (Level A) : "For all user interface components, the name and role can be programmatically determined." 

https://www.w3.org/TR/wai-aria-1.2/#row

### Reproducer

Use the following XHTML code:
```
<h:form id="frmTest">
    <pe:dynaForm id="dynaForm" value="#{testView.model}" var="data" 
                 pt:role="presentation"> <!-- Attempting to define layout table -->
        <pe:dynaFormControl type="input" for="txt">
             <p:inputText id="txt" value="#{data.value}"/>
        </pe:dynaFormControl>
    </pe:dynaForm>
</h:form>
```

The component renders the following HTML structure:

[primefaces-test-WCAG-DynaForm.zip](https://github.com/user-attachments/files/24035153/primefaces-test-WCAG-DynaForm.zip)

```
<table ... role="presentation">
    <tbody>
        <tr role="row"> <!-- Conflict here -->
            <td role="gridcell">...</td>
        </tr>
    </tbody>
</table>
```

### Expected behavior

May be the component should not force-render ARIA roles (role="row", role="gridcell") on standard HTML table elements ?

15.0.10

### JSF implementation

Mojarra

### JSF version

_No response_

### Browser(s)

_No response_

Vraken:


melloware:
Thanks I will handle porting this to 15 and 16

PR #2296, opened 2026-02-15
#2295: Search component - Support synonyms and update showcase

### LINKED ISSUE #2295: Search: Support synonyms

Support synonyms and update showcase

https://markjs.io/

> An object with synonyms. The key will be a synonym for the value and the value for the key. Example: "synonyms": {"one": "1"} will add the synonym "1" for "one" and vice versa

_No response_

_No response_

_No response_

### Additional context

_No response_

jxmai:
Resolve: #2295 

Local showcase demo: 

image

image

We can define synonyms map behind the scene as a map

```java
 private Map<String, String> synonyms = new HashMap<String, String>() {
        {
            put("one", "1");
            put("two", "2");
            put("three", "3");
            put("four", "4");
            put("five", "5");
        }
    };
```

### DIFF
--- a/core/src/main/java/org/primefaces/extensions/component/marktext/MarkText.java
+++ b/core/src/main/java/org/primefaces/extensions/component/marktext/MarkText.java
@@ -50,7 +50,6 @@ import com.google.gson.reflect.TypeToken;
 @ResourceDependency(library = "primefaces", name = "core.js")
 @ResourceDependency(library = "primefaces", name = "components.js")
 @ResourceDependency(library = Constants.LIBRARY, name = "mark/mark.js")
-@ResourceDependency(library = Constants.LIBRARY, name = "mark/marktext.js")
 public class MarkText extends MarkTextBase {
 
     public static final String COMPONENT_TYPE = "org.primefaces.extensions.component.MarkText";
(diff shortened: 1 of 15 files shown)


core/src/main/java/org/primefaces/extensions/component/marktext/MarkText.java
```diff
@@ -50,7 +50,6 @@
 @ResourceDependency(library = "primefaces", name = "core.js")
 @ResourceDependency(library = "primefaces", name = "components.js")
 @ResourceDependency(library = Constants.LIBRARY, name = "mark/mark.js")
-@ResourceDependency(library = Constants.LIBRARY, name = "mark/marktext.js")
```
jxmai, line 53: Found out this line is not needed. Removed

> WARNING: Resource referenced by resourceName mark/marktext.js and libraryName primefaces-extensions not found in call to ResourceHandler.createResource. It will be silenty ignored.

jxmai:
Strange that CI failed at a non-related error from another component.

> Error:  /home/runner/work/primefaces-extensions/primefaces-extensions/core/src/main/java/org/primefaces/extensions/component/tristatemanycheckbox/TriStateManyCheckboxRenderer.java:[52,70] type argument org.primefaces.extensions.component.tristatemanycheckbox.TriStateManyCheckbox is not within bounds of type-variable T

jxmai:
@melloware https://github.com/primefaces/primefaces/commit/0f22c0160134b313b3a1dca24173f4b81156275c

The issue was that SelectManyRenderer was declared with a generic type parameter <TriStateManyCheckbox>, but in the current PrimeFaces version (16.0.0-SNAPSHOT), SelectManyRenderer is not a generic class.

melloware:
Yes I plan to start working on the CDK upgrade soon.  They are still making update on the PF side as you can see.

PR #2323, opened 2026-02-21
CDK: FuzzySearch, GChart, Head

melloware:


core/src/main/java/org/primefaces/extensions/component/fuzzysearch/FuzzySearch.java
```diff
@@ -21,83 +21,64 @@
...
-    public Collection<String> getEventNames() {
-        return EVENT_NAMES;
+    public String getInputClientId() {
+        return getClientId(getFacesContext()) + "_input";
     }
 
     @Override
-    public String getDefaultEventName() {
-        return DEFAULT_EVENT;
+    public String getValidatableInputClientId() {
```
jxmai, line 62: Hello @melloware, I wonder where we convert this method from, since I don't see `getValidatableInputClientId` from the original codebase.

jxmai (approved):

core/src/main/java/org/primefaces/extensions/component/fuzzysearch/FuzzySearch.java
melloware: Yep this is new i implemented it the same way PF CascadeSelect which also extends UISelectOne so this makes this component more correct.

jxmai: Sounds great. That would be enhancement then. Thanks for the info!

PR #2325, opened 2026-02-21
#2324: CDK - Tooltip

### LINKED ISSUE #2324: CDK: Tooltip

jxmai:
Resolve: #2324 

Local showcase test drive:  http://localhost:8080/showcase/sections/tooltip/globalTooltips.jsf

image


core/src/main/java/org/primefaces/extensions/component/tooltip/TooltipBase.java
```diff
@@ -0,0 +1,117 @@
...
+    @Property(description = "Delay time for hiding the tooltip.", defaultValue = "0")
+    public abstract int getHideDelay();
+
+    @Property(description = "Hide effect function.", defaultValue = "fadeOut")
+    public abstract String getHideEffect();
+
+    @Property(description = "Duration for hide effect.", defaultValue = "500")
+    public abstract int getHideEffectLength();
+
+    @Property(description = "Id of the component to attach the tooltip.")
```
melloware, line 115: Can you check is the `for` required=true fot this component?

melloware (changes requested):

core/src/main/java/org/primefaces/extensions/component/tooltip/TooltipBase.java
jxmai: No, it wasn't required previously.

https://github.com/primefaces-extensions/primefaces-extensions/blob/846d16757ebdef8f61439e138fe2475f6aa72942/core/src/main/resources/META-INF/primefaces-extensions.taglib.xml#L4865-L4872

melloware: Nevermind confirmed its not required since you can use global mode!

melloware (approved):

PR #2329, opened 2026-02-21
#2328: CDK - Timer

### LINKED ISSUE #2328: CDK: Timer

jxmai:
Resolve: #2328 

Local showcase test drive: http://localhost:8080/showcase/sections/timer/jsapi.jsf

image


core/src/main/java/org/primefaces/extensions/component/timer/TimerBase.java
```diff
@@ -0,0 +1,141 @@
...
+import org.primefaces.component.api.StyleAware;
+import org.primefaces.component.api.Widget;
+
+/**
+ * <code>Timer</code> component base class.
+ *
+ * @author f.strazzullo
+ */
+@FacesComponentBase
+public abstract class TimerBase extends UIComponentBase implements Widget, AjaxSource, StyleAware {
```
melloware, line 39: Since you are implementing AjaxSource i think you can remove a lot of the props below they will come from AjaxSource generation?

core/src/main/java/org/primefaces/extensions/component/timer/Timer.java
```diff
@@ -113,284 +58,21 @@ public Locale calculateLocale() {
...
     @Override
     public boolean isPartialSubmitSet() {
-        return getStateHelper().get(PropertyKeys.partialSubmit) != null || getValueExpression("partialSubmit") != null;
+        return getStateHelper().get("partialSubmit") != null || getValueExpression("partialSubmit") != null;
     }
 
     @Override
     public boolean isResetValuesSet() {
-        return getStateHelper().get(PropertyKeys.resetValues) != null || getValueExpression("resetValues") != null;
+        return getStateHelper().get("resetValues") != null || getValueExpression("resetValues") != null;
```
melloware: I think you can still use PropertyKeys here as they are generated

core/src/main/java/org/primefaces/extensions/component/timer/Timer.java
```diff
@@ -113,284 +58,21 @@ public Locale calculateLocale() {
...
-    }
-
-    public void setIgnoreAutoUpdate(final boolean ignoreAutoUpdate) {
-        getStateHelper().put(PropertyKeys.ignoreAutoUpdate, ignoreAutoUpdate);
-    }
-
     @Override
     public boolean isPartialSubmitSet() {
-        return getStateHelper().get(PropertyKeys.partialSubmit) != null || getValueExpression("partialSubmit") != null;
+        return getStateHelper().get("partialSubmit") != null || getValueExpression("partialSubmit") != null;
```
melloware: Same i think you still want to use PropertyKeys here

melloware (changes requested):

core/src/main/java/org/primefaces/extensions/component/timer/TimerBase.java
jxmai: I see. Those are from the original implementation so I did not remove them. Let me review and clean those up if AajaxSource already implement them.

core/src/main/java/org/primefaces/extensions/component/timer/Timer.java
melloware: the first one is the enum the second is the name string like

```java
return getStateHelper().get(PropertyKeys.resetValues) != null || getValueExpression(PropertyKeys.resetValues.name()) != null;

core/src/main/java/org/primefaces/extensions/component/timer/TimerBase.java
jxmai: OK, https://github.com/primefaces/primefaces/blob/master/primefaces/src/main/java/org/primefaces/component/api/AjaxSource.java

will remove the redundant methods following the interfaces.

jxmai: https://github.com/primefaces-extensions/primefaces-extensions/pull/2329/changes/e0c42ed4005ba5f75b98ca067b6ce2bc8fe9c97e

core/src/main/java/org/primefaces/extensions/component/timer/Timer.java
jxmai: https://github.com/primefaces-extensions/primefaces-extensions/pull/2329/changes/6a07f12f96a87c9a42f99e81e1ee010381864031

core/src/main/java/org/primefaces/extensions/component/timer/Timer.java
jxmai: https://github.com/primefaces-extensions/primefaces-extensions/pull/2329/changes/6a07f12f96a87c9a42f99e81e1ee010381864031

jxmai:
Showcase is working as usual after the changes. Please take a look. Thanks @melloware .

melloware (approved):

PR #2330, opened 2026-02-21
CDK: Legend and Localized

melloware:


monacoeditor/src/main/js/package-lock.json
jxmai, line 1: wondering monacoeditor generates diff on build every time, If the change is intended?

melloware: Yep I need to investigate that.  @blutorange is the expert there so may need him to weigh in!

Issue #2331, opened 2026-02-21
Monacoeditor: package-lock.json modified on Every Build & Windows cannot clean up target/git Directory


jxmai:
### Describe the bug

On Primefaces extensions build,

there's 2 issues on Monacoeditor.

1. package-lock.json modified on every build, and we experienced the pain on avoid / accidentally commit the file.

2. On windows, every Monacoeditor build will lock the file `monacoeditor/target/git/vscode-loc/.git/objects/pack/*.idx`, that we have to manually clean up.

### Reproducer

`mvn clean install` twice would see the problem.

### Expected behavior

There shouldn't be such a painful process to build project that we have to mannually avoid / adjust monacoeditor. This is epecially haunting on active development.

The first issue is because we use `npm install` instead of `npm ci`. Please read through: https://support.deploybot.com/build-tools/why-developers-should-use-npm-ci-instead-of-npm-install-and-its-benefits that explains the differences.

Conclusion:

> In conclusion, npm ci is a command that developers should consider using instead of npm install for their Node.js projects. It provides consistency, speed, predictability, and automation benefits, making it a valuable tool for any development team. So, if you haven't already, give npm ci a try and see how it can improve your workflow.

The second problem can be addressed via: https://www.npmjs.com/package/rimraf.

The root cause is: The build clones Microsoft's vscode-loc repository for locale generation. Git's internal files create locks on Windows that Maven's clean plugin cannot handle

What we can do is to add this clean up cmd before maven execute on maven clean to assist clean up.

### PrimeFaces Extensions version

PF 16.0.0

### JSF implementation

None

melloware:
Added @blutorange on this

jxmai:
I submitted a draft: https://github.com/primefaces-extensions/primefaces-extensions/pull/2332

Please feel free to take a look. Ideally it would be great if the file not being locked in the first place. but it looks like maven couldn't hold the lock for the generated file, which requires https://www.npmjs.com/package/rimraf to clean up. 

npm ci changes on the other hand should be standard changes as we shouldn't use npm install in this context.

jxmai:
Currently building on windows `mvn clean install` failed every second time unless stepping into the file explorer / use some cmd such as rmdir manually to delete the target folder in Monacoeditor.

```bash
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-clean-plugin:3.5.0:clean (default-clean) on project resources-monacoeditor:
```

PR #2332, opened 2026-02-21
#2331: prevent package-lock.json modifications and Windows cleanup failure in monacoeditor build

jxmai:
#2331

### DIFF
--- a/monacoeditor/pom.xml
+++ b/monacoeditor/pom.xml
@@ -48,7 +48,14 @@
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-clean-plugin</artifactId>
                 <configuration>
+                    <excludeDefaultDirectories>true</excludeDefaultDirectories>
                     <filesets>
+                        <fileset>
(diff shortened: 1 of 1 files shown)


melloware (approved):

melloware:
Seems reasonable to me!

jxmai:
I ran on local, mvn clean install consequencely work after the changes. The reason we need to exclude a maven folder because of the clean timing, if maven clean first, the locked file failed so maven need to skip that folder letting rimraf to clean.

jxmai:
The other option is to move that lock file out of the target folder such as creating a hidden cache, but the downside is that it would be left of out of the clean up life cycle, so using npm tool to clean and coordinate with maven could be a better choice.

Most importantly, rimraf is cross platform so it will work for other OSes.

jxmai:
The best solution that I can think of is those files shouldn't be locked in the first place inside the target folder generated by npm, but it does not look like within our control.

melloware:
Agreed

PR #2351, opened 2026-02-24
#2350: Spotless: Enforce endWithNewLine on Java source code

### LINKED ISSUE #2350: Spotless: Enforce endWithNewLine on Java source code
I noticed some of PRs I sent contains `fragment`, https://github.com/primefaces-extensions/primefaces-extensions/pull/2325/changes, meaning end of file does not end with new line, but they slipped through. It wouldn't break the functionality but it would be nice we clean those up.

---

https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline

> 3.206 Line
A sequence of zero or more non- <newline> characters plus a terminating <newline> character.

jxmai:
Resolve #2350


jxmai:
Let me think about if the config can be simplfiied even further.

melloware (approved):

melloware:
Sure thing.

jxmai:
I am reviewing whether I should add 

```java
<includes>
                                <include>src/main/java/**/*.java</include>
                                <include>src/test/java/**/*.java</include>
                            </includes>
```
or not, will let you know shortly!

jxmai:
Confirmed 

```xml
<includes>
                                <include>src/main/java/**/*.java</include>
                                <include>src/test/java/**/*.java</include>
                            </includes>
```

is not needed. 

In the end, just a single line change, and spotless could catch all the issues.

jxmai:
@melloware All good 👍

PR #2451, opened 2026-04-13
SunEditor 2.47.10

melloware:
https://github.com/JiHong88/suneditor/releases/tag/2.47.10


florian-dahlmanns-inform:
Hi @melloware ,
is PrimeFaces affected by the reported vulnerability within the SunEditor or is the vulnerability not exploitable due to some kind of the JS library being wrapped by PrimeFaces? If it is exploitable, will there be also a backport for PrimeFaces Extensions 13 and or even 11(if PFE 13/11 is affected, but I guess so).

And I'm wondering why there is no separate GitHub issue, just the two pull requests. A ticket would have probably answered these questions already and would have some sort of "official character" so others can directly see what's going on.

Thanks :)

melloware:
i didn't really study the exact vulnerability or whether PFE is affected.  I think SunEditor only goes back to PFE13 so 13, 14, 15 would need releases with this fix.  If you want to create a ticket so we can track it that would be great.

melloware:
i released 13.0.20 with this fix

florian-dahlmanns-inform:
Hi @melloware, you are right, SunEditor was introduced in PFE 12, so PFE 11 cannot be affected. Thanks for your information here. But I'm wondering where the release "13.0.20" is... I cannot see it on Maven Centrel yet...

melloware:
13.0.21 is on Central here: https://repo1.maven.org/maven2/org/primefaces/extensions/primefaces-extensions/13.0.21/

Issue #2436, opened 2026-03-31
MeterGroup: new component


NicolaIsotta:
### Describe the feature you would like to see added

I just noticed a new MeterGroup component in PrimeNG -> https://primeng.org/metergroup

### Additional context

Currently working on implementing the component with the help of Google Antigravity

melloware:
would love this. Since this is a core PF component you might be able to do this as a PF Core component. I think its similar to ChronoLine etc https://showcase.primefaces.org/ui/data/chronoline.xhtml?jfwid=2f907

which is jhust CSS and HTML mostly.

NicolaIsotta:
Needed a few tweaks, but it's looking fine now:
<img width="1220" height="900" alt="Image" src="https://github.com/user-attachments/assets/7d316050-477d-467c-80d5-f5ff981a2da7" />

melloware:
nice!

PR #2462, opened 2026-04-18
#2436: update version spec

jxmai:
#2436  Documentation update


core/src/main/java/org/primefaces/extensions/model/marktext/MarkPosition.java
```diff
@@ -27,7 +27,7 @@
  * Model class for marktext positions.
  *
  * @author jxmai
- * @since 16.0.0
+ * @since 15.0.13
```
jxmai, line 30: This isn't related to MeterGroup, but I fixed it in this commit when I noticed it since it's harmless.

melloware (approved):

PR #2482, opened 2026-05-01
#1740: FluidGrid - dynamically replaced items shows the same label on form submitted

### LINKED ISSUE #1740: FluidGrid : when change value, labels of components lost your label and all components use the label last component .
### Describe the bug

When the FluidGrid component is used in the form of a dynamic form, and this form has components that are required and each one with its own label, when the form is sent and the validations of the required components are skipped, they always have the same label as the last component that was rendered.
This only happens when the values of the value attribute of the fluidgrid are replaced by other elements, when the component is updated, when the form is resubmitted, the validation messages of the components are displayed and in this case, all the errors show the same validation error, pointing to the last component that was rendered.
The only way to restore this behaviour is to refresh the page, but again this will only work for the first components that are rendered, if they are replaced by others, the error occurs and it is no longer possible to fix it.

It has been tested resetting to null the value and then doing an update from the bean to stop rendering the component and then set the new values to the list and update again to render the component again with the new values, it has been tested with <c:if tags to render the content that would be the fluidgrid, cleaning all the elements of the value and setting new ones. 
It has also been tried to update the form or the component itself, the result is the same, the component label is not correctly assigned when validations are processed.

Also tested with the latest version of primefaces (14.0.3) and primefaces extension (14.0.6) and the same thing happens

Gif with bug:

bugPrimefacesExtensions

### Reproducer

It can be reproduced with the same showcase case, adding a selectOneRadio for example, with a listener that sets the new list of dynamic values and updating the component again. I attach the project where the problem is reproduced.

### Expected behavior

When the fluidgrid is updated, all components' labels must be set correctly again, not all components assigned the same label.

Example code:
[primefaces-test-7.0.zip](https://github.com/user-attachments/files/17601097/primefaces-test-7.0.zip)

7.0.3

### JSF implementation

Mojarra

### JSF version

2.2.20

### Browser(s)

All

jxmai:
Resolve #1740 

---

The problem was that when fluidGrid items are dynamically replaced, such as via AJAX, and the form is submitted with the validation errors, all the valiation message shows the same label (The label of the last rendered item), which is clearly bug.

In a deeper dive, there's actually two bugs in [AbstractDynamicData.java](https://github.com/primefaces-extensions/primefaces-extensions/blob/master/core/src/main/java/org/primefaces/extensions/component/api/AbstractDynamicData.java)

### Issue 1: Wrong var in state lookup

https://github.com/primefaces-extensions/primefaces-extensions/blob/63b4a93aaf1d8d567190335d178e71b2f1f1b5a8/core/src/main/java/org/primefaces/extensions/component/api/AbstractDynamicData.java#L545

It uses fluidgrid component's owned cache clientId, 

https://github.com/primefaces-extensions/primefaces-extensions/blob/63b4a93aaf1d8d567190335d178e71b2f1f1b5a8/core/src/main/java/org/primefaces/extensions/component/api/AbstractDynamicData.java#L67

It sets once at here, https://github.com/primefaces-extensions/primefaces-extensions/blob/63b4a93aaf1d8d567190335d178e71b2f1f1b5a8/core/src/main/java/org/primefaces/extensions/component/api/AbstractDynamicData.java#L167-L169

It is the parent's component own cached clientId, as a result `saved.get(clientId)` looks up `form:grid` -> always null -> create new state, which also means the EditableValueHolder creates an empty SaveEditableState instead of updating the one that was stored from the previous save/restore cycle. This is the reason caused label value and other state to be lost rather than preserved across row iterations.

If we change it to `saved.get(id)`, it would look up `form:grid:id` -> finds existing state -> updates

### Issue 2: Label restoration destroys ValueExpression

https://github.com/primefaces-extensions/primefaces-extensions/blob/63b4a93aaf1d8d567190335d178e71b2f1f1b5a8/core/src/main/java/org/primefaces/extensions/component/api/AbstractDynamicData.java#L603-L605

restoreDescendantState called `.getAttributes().put(Attrs.LABEL, state.getLabelValue());`, which replaced the value expression such as `data.label` with a hardcoded string.  Once replaced, all items of the same type see the same literal instead of re-evaluate the the EL per row, so we need to guard it with `component.getValueExpression(Attrs.LABEL) == null`

---

New showcase demo:

http://localhost:8080/showcase/sections/fluidgrid/dynamiclabels.jsf

image

image

### DIFF
--- a/core/src/main/java/org/primefaces/extensions/component/api/AbstractDynamicData.java
+++ b/core/src/main/java/org/primefaces/extensions/component/api/AbstractDynamicData.java
@@ -542,7 +542,7 @@ public abstract class AbstractDynamicData extends UIComponentBase implements Nam
             }
 
             if (state == null) {
-                state = saved.get(clientId);
+                state = saved.get(id);
 
                 if (state == null) {
                     state = new SavedEditableValueState();
@@ -600,7 +600,7 @@ public abstract class AbstractDynamicData extends UIComponentBase implements Nam
             input.setValid(state.isValid());
             input.setSubmittedValue(state.getSubmittedValue());
             input.setLocalValueSet(state.isLocalValueSet());
-            if (state.getLabelValue() != null) {
+            if (state.getLabelValue() != null && component.getValueExpression(Attrs.LABEL) == null) {
                 ((UIComponent) input).getAttributes().put(Attrs.LABEL, state.getLabelValue());
             }
 
--- a/showcase/src/main/java/org/primefaces/extensions/showcase/controller/fluidgrid/FluidGridDynamicLabelsController.java
+++ b/showcase/src/main/java/org/primefaces/extensions/showcase/controller/fluidgrid/FluidGridDynamicLabelsController.java
@@ -1,0 +1,116 @@
+/*
+ * Copyright (c) 2011-2026 PrimeFaces Extensions
+ *
+ *  Permission is hereby granted, free of charge, to any person obtaining a copy
+ *  of this software and associated documentation files (the "Software"), to deal
+ *  in the Software without restriction, including without limitation the rights
+ *  to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ *  copies of the Software, and to permit persons to whom the Software is
+ *  furnished to do so, subject to the following conditions:
+ *
+ *  The above copyright notice and this permission notice shall be included in
+ *  all copies or substantial portions of the Software.
+ *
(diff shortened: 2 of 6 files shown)


jxmai:
It looks many comments for some reason. I will re-review.

jxmai:
Nevermind, those are the comments on the showcase. The new showcase is to demo the usecase. Since the changes is small but necessarily pervasive, need showcase demo as testing.

melloware (approved):

Issue #2571, opened 2026-05-30
Kanban: board text unreadable in dark themes (e.g. vela-blue)


jxmai:
### Describe the feature you would like to see added

When using a dark PrimeFaces theme like vela-blue, the kanban board becomes unreadable because the CSS uses hardcoded colors instead of theme CSS variables.

https://primefaces-extensions-showcase-55pq8.ondigitalocean.app/sections/kanban/basicUsage.jsf

<img width="923" height="417" alt="Image" src="https://github.com/user-attachments/assets/83da174f-dec0-4651-b23a-ae4d69e8fe0d" />

jxmai:
The style was hardcoded in the original Kanban library, to avoid directly changing third party source, we will override from primefaces extensions.

PR #2598, opened 2026-06-12
#1066: CommandPalette: New Component

### LINKED ISSUE #1066: CommandPalette: New Web Component

Recently I had to create a Web Component for something I needed because there is no library (vanilla or jquery) available for it and I was about to use it on one of our PF apps, so I thought maybe it could be added to PFE if others find it useful. @melloware suggested I create an issue so we can discuss about it. Basically, the idea is having a more detailed and filterable context menu. This is heavily inspired by the blueprint context menu in unreal engine, if you're not familiar with it, it looks like this:

https://user-images.githubusercontent.com/32560510/219885462-f82cae0c-ecef-420c-a1ef-7ca1c39a1eb0.mp4

This is my version of it: 

https://user-images.githubusercontent.com/32560510/219885687-35ab79ba-cc81-43a0-9129-8d04ddc79360.mp4

This component is part of an existing Web Component collection of mine where I create self contained Web Components with zero external dependencies or references to any external files, with the goal being simplicity, reliance on browser web standards, and being framework agnostic.  As a result, the entire component+styling is 1 single .js file.

_No response_

_No response_

_No response_

### Additional context

Link to the component code: https://github.com/newk5/stak-components/blob/master/components/st-grouplist.js
Link to showcase: https://newk5.github.io/stak-components/

Usage Example: 
```html
<script src="js/stak-grouplist.js" type="module"></script>

<div id="test" style="width: 100px; height:100px; border:solid 2px red; padding: 5px;">
        <br/><br/>
        right click here
 </div>

 <st-grouplist id="groups" for="#test" label="Nodes" placeholder="Search nodes...">
        <option group="Node1" value="1">Create Node1 Action1</option>
        <option group="Node1" value="2">Create Node1 Action2</option>
        <option group="Node2" value="3">Create Node2 Action1</option>
        <option group="Node2" value="4">Create Node2 Action2</option>
 </st-grouplist>
  
  <script>
        document.addEventListener("DOMContentLoaded", () => {
            let groups = document.querySelector("#groups");
            groups.onSelect(groupOption => {
                console.log(`You selected option "${groupOption.label}" with value: "${groupOption.value}" from group: "${groupOption.name}" `)
            });
        });
  </script>
```
For a small title tooltip, each `<option>` can also have a groupTitle="" and optionTitle=""

Since this is a webcomponent that uses a shadow dom, the .css thats being applied inside the component will not affect anything outside it, and as a result, css applied outside the component will also not affect it. There's 3 different ways to customize the styling of the component, see [here](https://github.com/newk5/stak-components/wiki#styling)

Also I'm not sure if I like the name "grouplist" too much, this is indeed a list of groups with each having a set of options, but it's also a context menu so I dont know what else can I name it.

jxmai:
Resolve: #1066 

This is to address the main goal from the ticket: 

> Basically, the idea is having a more detailed and filterable context menu.

There's no OOTB solution, but the widget JS is adapted from [st-grouplist.js](https://github.com/newk5/stak-components/blob/master/components/st-grouplist.js) in the [newk5/stak-components](https://github.com/newk5/stak-components) customized as component.

Demo: http://localhost:8080/showcase/sections/commandPalette/basicUsage.jsf

image

image


melloware:
Adding @newk5

newk5:
Looks great! Thanks for implementing it.


About the name, do we keep it as "CommandPalette"? I'm fine with the name if you guys are, I was just thinking if we ever add a more traditional CommandPalette to PFE in the future like the ones @melloware mentioned in #2598, this one already having that name and looking similar to right click context menus, might cause some confusion. What do you think?

melloware:
yeah what differentiates this from ContextMenu in PF?

newk5:
The main differences would be that this has a filter input that allows searching all sub levels. It was made for when there's a large number of items. Being able to filter for specific items or sections comes in handy when you have large number of them. And each item must belong to to a "section". The idea was to group common items in sections, for example "File operations" section with children like: cut, copy, replace, upload, download etc... 

You can see where I got the idea from in the first video I posted in #1066 ( I linked the wrong issue in my previous reply, its #1066 not  #2598)

jxmai:
Looking at the conversations, there are two main points to address, so let's settle them:

1. Naming convention: CommandPalette or GroupList.
2. How it differs from PrimeFaces context menu.

I recommend CommandPalette over GroupList. That also clarifies how it differs from PrimeFaces context menu when we add new features to this component. In the demo, the new component was triggered by the contextmenu event, but that’s just the demo,  we can potential scale it to be triggered by other events (for example, keyboard events), which would align it more closely with the CommandPalette definition.

https://uxpatterns.dev/patterns/advanced/command-palette

> A Command Palette pattern helps teams create a reliable way to find and run commands, destinations, and recent items from a single keyboard-first surface. It is most useful when teams need editor and workspace commands.

Leaving it as Command Palette gives a concrete goal to develop toward. At first it shows additional features such as filtering within commands, but it could potentially do more.

melloware:
Yes we should have `<p:hotKey>` example where its trigger by a keyboard combination.

newk5:
Alright, sounds good to me, we can keep the CommandPalette name and in the future expand on it. As for why I chose to use the right click, it was because I was trying to emulate the way it works in the unreal engine blueprint editor, which is that the list of actions shows at the location where you right click with the mouse cursor and allows you to put nodes at that location like this:
UnrealEditor_0I5J0jBK8k

So the option/item you choose, is then put at the location where you did your right mouse click. But yea as you said, I see no reason to limit it to right click only, we could in the future allow different ways to trigger it

jxmai:
Updated: http://localhost:8080/showcase/sections/commandPalette/hotkey.jsf

image

Showcasing https://showcase.primefaces.org/ui/misc/hotkey.xhtml integration with CommandPalette.

melloware:
@newk5 you can check it out after it deploys to the showcase here: https://primefaces-extensions-showcase-55pq8.ondigitalocean.app/

newk5:
Looking great, just 1 small thing I noticed, it seems on some dark themes the color of some stuff is hard to see, some examples:
- Using arya-blue
image

- Using vela-blue
image


- Using luna-blue
image


- Using nova-light
image

jxmai:
Sounds good. No problem. That should be similar to a theme tweak I did in the past for a different component.: https://github.com/primefaces-extensions/primefaces-extensions/issues/2571

I will review and enhance the theme handling after backporting the main code to PF 15.

PR #2704, opened 2026-08-21
harden: an expression is built with a dynamic value in...

anupamme:
## Summary
Harden input handling in `core/src/main/java/org/primefaces/extensions/application/PrimeFacesResourceProcessor.java` (flagged by semgrep).

## Vulnerability
| Field | Value |
|-------|-------|
| **ID** | java.lang.security.audit.el-injection.el-injection |
| **Severity** | HIGH |
| **Scanner** | semgrep |
| **Rule** | `java.lang.security.audit.el-injection.el-injection` |
| **File** | `core/src/main/java/org/primefaces/extensions/application/PrimeFacesResourceProcessor.java:72` |
| **Assessment** | Defensive hardening |

**Description**: An expression is built with a dynamic value. The source of the value(s) should be verified to avoid that unfiltered values fall into this risky code evaluation.

## Threat Model Context

This is a Java service - vulnerabilities in servlets/controllers are remotely exploitable.

## Changes
- `core/src/main/java/org/primefaces/extensions/application/PrimeFacesResourceProcessor.java`

## Behavior Preservation
The change is scoped to 1 file on the vulnerable path; it only tightens handling of untrusted input and leaves valid inputs unaffected.

---
*This patch removes an exploit primitive — a code pattern that, while not independently exploitable today, could be chained with other weaknesses by automated exploit-development tooling. Proactive removal of such primitives raises the bar against increasingly capable automated attack tools.*

---
*Automated security fix by [OrbisAI Security](https://orbisappsec.com)*

### DIFF
--- a/core/src/main/java/org/primefaces/extensions/application/PrimeFacesResourceProcessor.java
+++ b/core/src/main/java/org/primefaces/extensions/application/PrimeFacesResourceProcessor.java
@@ -23,9 +23,6 @@ package org.primefaces.extensions.application;
 
 import java.util.Map;
 
-import jakarta.el.ELContext;
-import jakarta.el.ExpressionFactory;
-import jakarta.el.ValueExpression;
 import jakarta.faces.context.FacesContext;
 import jakarta.faces.event.PhaseEvent;
 import jakarta.faces.event.PhaseId;
@@ -81,10 +78,7 @@ public class PrimeFacesResourceProcessor implements PhaseListener {
 
         if (themeParamValue != null) {
             if (themeParamValue.startsWith("#{") || themeParamValue.startsWith("${")) {
-                ELContext elContext = context.getELContext();
-                ExpressionFactory expressionFactory = context.getApplication().getExpressionFactory();
-                ValueExpression ve = expressionFactory.createValueExpression(elContext, themeParamValue, String.class);
-                theme = (String) ve.getValue(elContext);
+                theme = context.getApplication().evaluateExpressionGet(context, themeParamValue, String.class);
             }
             else {
                 theme = themeParamValue;
--- a/core/src/main/java/org/primefaces/extensions/application/PrimeFacesResourceProcessor.java
+++ b/core/src/main/java/org/primefaces/extensions/application/PrimeFacesResourceProcessor.java
@@ -80,11 +80,15 @@ public class PrimeFacesResourceProcessor implements PhaseListener {
         String themeParamValue = applicationContext.getConfig().getTheme();
 
         if (themeParamValue != null) {
-            ELContext elContext = context.getELContext();
-            ExpressionFactory expressionFactory = context.getApplication().getExpressionFactory();
-            ValueExpression ve = expressionFactory.createValueExpression(elContext, themeParamValue, String.class);
-
-            theme = (String) ve.getValue(elContext);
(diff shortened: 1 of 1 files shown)


melloware (approved):

melloware:
@anupamme  did you check the PrimeFaces codebase also for any similar issues?

melloware:
@tandraschko  wasn't sure if this was necessary i thought this was defended in the Faces implementation itself ?

i noticed PF does it a little differently

```java
String themeParamValue = applicationContext.getConfig().getTheme();

if (themeParamValue != null) {
   theme = context.getApplication().evaluateExpressionGet(context, themeParamValue, String.class);
}
```

anupamme:
> @anupamme did you check the PrimeFaces codebase also for any similar issues?

Yes, there are a few vulnerabilities, but they are MEDIUM/LOW severity. If you wish, I can send you a report.

melloware:
Yes please send the report

anupamme:
> Yes please send the report

sent to your gmail id.

piolot:
@melloware: I cannot see that you prevent the el-injection, you are just checking that the configured value is an el and then you still execute it. The only true fix here would be to avoid el completly in the web.xml. Still I do not understand why an el in the web.xml is dangerous. If you have access to the web.xml and therefor to the WAR file nobody could stop you from putting malicious code to the WAR file.

melloware:
@piolot he mentions it is NOT really exploitable but "could be" one day.  This is just a check that it is an EL expression instead of blindly calling EL.   

But you are exactly right this is more defensive code than an actual exploit.

tandraschko:
tBH i would check If the EL impl already does this - im sure it does

piolot:
@melloware : I still do not see that the defensive code has anything to do with the mentioned el-injection problem. I would think that the JSF implementation already checks if the string contains an el and otherwise just returns the string.

tandraschko:
yeah
propably JSF and EL impl already does it

and still, this code change does not prevent anything

jxmai:
@melloware  This looks like one of the classic examples of static security analysis false positive, that we did not have vunlerability in the first place.

Let's break it down, the value comes from primefaces.THEME with context param in web.xml. Only someone who can edit web.xml can control it, and anyone can do that already owns the entire WAR, so no privledge boundary is crossed.

any EL injection needs to have user input that touches the code path, in this case, nothing from the request can touches this code path.

and the "fix" actually leaves the regression like `prefix-{bean.value}` now pass through unevaluated. It is not fixing anything but break compatibiliy.

anupamme:
Thanks, that’s a fair point. I agree I focused too much on the EL sink identified by static analysis and not enough on the actual attacker-controlled data flow.

Given that primefaces.THEME is coming from web.xml, I agree that this does not appear to cross a privilege boundary in the usual deployment model, and the finding may indeed be a false positive.

I also hadn’t considered the compatibility implication of only evaluating values beginning with #{/${; in particular, expressions embedded in a larger theme value such as prefix-#{bean.value} would now be treated literally.

Since the PR is already merged, I can take a closer look at the existing supported behaviour and the impact of this change. If the security benefit isn’t meaningful and the compatibility regression is real, I can raise another PR which reverts or adjusts the change.

melloware:
I would just revert.

PR #2719, opened 2026-08-31
#2718: new option: outsideClickDeselectsBoolean

### LINKED ISSUE #2718: Sheet: highlighting of row removed
### Describe the bug

When clicking somewhere outside the sheet the css classes `current highlight` are removed.

- select a row → highlighted
- click outside the sheet → not highlighted

### Reproducer

Can be reproduced here: https://extensions.primefaces.org/sections/sheet/validation.jsf

### Expected behavior

It would be more logical to only remove those classes when users select another row.

15.0.17

### JSF implementation

None

### JSF version

_No response_

### Browser(s)

_No response_

jxmai:
Resolve: #2718 

https://josepsanzcamp.github.io/handsontable-6.2.2-docs/Options.html#outsideClickDeselects

> If true, mouse click outside the grid will deselect the current selection. Can be a function that takes the
click event target and returns a boolean.

> Default Value: true
Example
// don't clear current selection when mouse click was outside the grid
outsideClickDeselects: false,

> // or
outsideClickDeselects: function(event) {
  return false;
}

### DIFF
--- a/core/src/main/java/org/primefaces/extensions/component/sheet/Sheet.java
+++ b/core/src/main/java/org/primefaces/extensions/component/sheet/Sheet.java
@@ -856,7 +856,10 @@ public class Sheet extends SheetBaseImpl {
         }
 
         if (restoredSortedList == null) {
-            getFilteredValue().clear();
+            final List<Object> filteredValue = getFilteredValue();
+            if (filteredValue != null) {
+                filteredValue.clear();
+            }
         }
         else {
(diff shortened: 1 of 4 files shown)


jxmai:
This part could cause an NPE, fix all of them together.

```java
if (restoredSortedList == null) {
            getFilteredValue().clear();
```

melloware (approved):

